PR #1, opened 2023-06-19
Shared GitHub Actions for CNB Releases

colincasey:
Initial commit of shared workflows and actions for release automation tasks for Cloud-native Buildpacks.

[W-13168458](https://gus.lightning.force.com/a07EE00001QzelrYAB)


Malax (approved):
This looks good so far! :) I can imagine that it will change a bit in the next couple of days when we roll this out, so there is no need to nitpick about details I think. Love that the CI for this repo itself looks solid! The only thing we should add (either in the PR or as a follow-up) is an example how to use the shared workflow. Currently, only the actions themselves are documented. Thank you for working on this! :)

colincasey:
yes, i was hesitant around the shared workflows (the CNB release one isn't even in this commit) because i think those will require minor modifications as we go.

src/main.rs
```diff
@@ -0,0 +1,55 @@
+use crate::commands::generate_buildpack_matrix::command::GenerateBuildpackMatrixArgs;
```
edmorley, line 1: Could you enable the usual lints additional lints we use for our Rust repos? :-)

```
#![warn(clippy::pedantic)]
#![warn(unused_crate_dependencies)]
```

dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "monthly"
```
edmorley, line 14: ❤️

Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+    "derive",
+    "error-context",
+    "help",
+    "std",
+    "usage",
+] }
+indexmap = "1.9.3"
+lazy_static = "1.4.0"
+libcnb-data = { git = "https://github.com/heroku/libcnb.rs", branch = "feature/W-12731729_libcnb_package" }
+libcnb-package = { git = "https://github.com/heroku/libcnb.rs", branch = "feature/W-12731729_libcnb_package" }
```
edmorley, line 26: I presume this is on your list already, but once there is a new libcnb.rs release, the Git dependency here should be changed to crates.io

edmorley (commented):
Thank you for this!

src/main.rs
edmorley: Filed #36

Cargo.toml
edmorley: Resolved in d470f06179e81039d730289e69a3290f9620bdd9

PR #18, opened 2023-06-21
Add --declarations-starting-version arg to prepare-release

colincasey:
The optional argument `--declarations-starting-version` has been added to the `actions prepare-release` command to limit the declarations created when rebuilding the `CHANGELOG.md` file into the Keep A Changelog format.

Buildpacks that are only now normalizing on fixed versions and release tags (e.g.; [`buildpacks-jvm`](https://github.com/heroku/buildpacks-jvm), [`buildpacks-nodejs`](https://github.com/heroku/buildpacks-nodejs)) can use this to prevent invalid declaration links from being generated.


Malax (approved):
This sneakily also fixes another issue without calling that out in the PR - took me a sec to separate the two things. I caught you. ;-) LGTM!

PR #28, opened 2023-06-22
Added action for checking if a buildpack is already published to the CNB registry

colincasey:


.github/workflows/_buildpacks-release-publish-cnb-registry.yml
```diff
@@ -27,17 +27,23 @@ jobs:
...
-      - name: Check if version is already in the registry
-        id: check
-        run: |
-          export URL="https://registry.buildpacks.io/api/v1/buildpacks/${{ inputs.buildpack_id }}/${{ steps.buildpack.outputs.buildpack_version }}"
-          export EXISTS=$(if [ "$( curl -s -o /dev/null -I -w "%{http_code}" "${URL}")" = "200" ]; then echo 'true'; else echo 'false'; fi)
-          echo "published_to_cnb_registry=${EXISTS}" >> $GITHUB_OUTPUT
       - name: Calculate the buildpack image digest
         id: digest
         run: echo "value=$(crane digest ${{ steps.buildpack.outputs.docker_repository }}:${{ steps.buildpack.outputs.buildpack_version }})" >> "$GITHUB_OUTPUT"
+      - name: Checkout buildpack registry index
```
edmorley, line 33: What are your thoughts on using the buildpack registry API instead of cloning the index (which is an internal implementation detail of the registry).

eg:
https://registry.buildpacks.io/api/v1/buildpacks/heroku/nodejs
https://registry.buildpacks.io/api/v1/buildpacks/heroku/nodejs/0.6.7

colincasey: that's what we were doing [here](https://github.com/heroku/languages-github-actions/blob/main/.github/workflows/_buildpacks-release-publish-cnb-registry.yml#L33)  @edmorley but sometimes the registry API isn't updated immediately or responses are cached for some amount of time.

edmorley: There's a delay at the moment only due to a bug/limitation which is about to be fixed:
https://github.com/buildpacks/registry-api/issues/113

edmorley: It seems there are quite a few states the registration process could be in at the time a duplicate job is retriggered on our side. ie:
1. The "register new buildpack version issue" may not have even been created
2. The "register new buildpack version issue" could exist, be open, but the bot that listens for them hasn't actioned it yet (eg gotten stuck, GitHub Actions runner backlog)
3. The "register new buildpack version issue" could have been closed out by the bot but not actioned (eg some transient error)
4. The "register new buildpack version issue" could have been actioned (commit added to `main`) and closed out, but the API indexer not yet picked up the change.

I guess the question is what stages need to be protected against?

Malax: My 5c:

It's not really about "protection" I think. It's more about convenience when using the release pipeline. If you retrigger (parts) of the pipeline and it will fail again because the check is unreliable (due to caching, late updates) it would suck. In this specific case where the updater upstream is broken, we would not even get out of the broken state anymore (in which case it is more about "protection" I guess).

The workaround of using the same canonical data as the check upstream (the GHA that handles the GH issues) we can check if that upstream action will most likely fail or not - regardless of other uses of the data. Our only concern here (correct me if I'm wrong) is "will the upstream GH issue be handled successfully?". 

Even without the ongoing issue with the registry index updates, this new solution should be much more reliable and the amount of code for it isn't too bad IMO. Maybe we can improve it by only doing a sparse git clone with only the path that we actually need, but that can be a follow-up.

joshwlewis: Alternatively, you can skip the clone completely and use GitHub's API. For example, https://api.github.com/repos/buildpacks/registry-index/contents/no/de/heroku_nodejs returns

```json
{
  "name": "heroku_nodejs",
  "path": "no/de/heroku_nodejs",
  "sha": "f0c1d7e93277ce3cb8ad3e27614d380d41102b07",
  "size": 6845,
  "url": "https://api.github.com/repos/buildpacks/registry-index/contents/no/de/heroku_nodejs?ref=main",
  "html_url": "https://github.com/buildpacks/registry-index/blob/main/no/de/heroku_nodejs",
  "git_url": "https://api.github.com/repos/buildpacks/registry-index/git/blobs/f0c1d7e93277ce3cb8ad3e27614d380d41102b07",
  "download_url": "https://raw.githubusercontent.com/buildpacks/registry-index/main/no/de/heroku_nodejs",
  "type": "file",
  "content": "encodedebase64contentwashere=\n",
  "encoding": "base64",
  "_links": {
    "self": "https://api.github.com/repos/buildpacks/registry-index/contents/no/de/heroku_nodejs?ref=main",
    "git": "https://api.github.com/repos/buildpacks/registry-index/git/blobs/f0c1d7e93277ce3cb8ad3e27614d380d41102b07",
    "html": "https://github.com/buildpacks/registry-index/blob/main/no/de/heroku_nodejs"
  }
}
```

You can decode the "content" field into json, or download the file directly from `https://raw.githubusercontent.com/buildpacks/registry-index/main/no/de/heroku_nodejs`.

edmorley: I like the idea of using `raw.githubusercontent.com` over git cloning, given that:
- as the index grows the git clone times will eventually slow down vs downloading the file directly
- a lot of other tools/instructions on the web fetch from`raw.githubusercontent.com` when downloading eg bootstrap scripts etc, so it's not an uncommon pattern to use

colincasey: now that the underlying issue has been fixed with the registry api indexing i think we should close this.  the state that @Malax and i encountered with the index not updating for hours made the added complexity of this PR seem worthwhile but if we really only have to worry about a small delay with re-indexing on the api then it's unlikely we'll be severely blocked like this again.

and if the registry api does start to become a problem in the future we can revisit with the approach @joshwlewis suggested.

PR #35, opened 2023-06-26
Don't allow republishing of the shared github actions to the same release tag

### LINKED ISSUE #13: Publishing does not block when version already exists

GUS-W-13658882.

colincasey:
The current build process will overwrite any existing release version without warning.  The following steps have been added to prevent this:
* Check if there are any changes made to binary code
  * If yes, check if the tag exists for the version in `Cargo.toml` and, if it does, error with a message to bump the version
* Only publish the binary when the binary code has changed

Fixes #13.
[GUS-W-13658882](https://gus.lightning.force.com/a07EE00001Ul6LdYAJ)

### DIFF
--- a/.github/workflows/build-test-release.yml
+++ b/.github/workflows/build-test-release.yml
@@ -64,25 +64,13 @@ jobs:
             exit 1
           fi
 
-  check-binary:
+  check-tag:
     name: Check Binary
     runs-on: ubuntu-latest
-    outputs:
-      needs_to_be_published: ${{ steps.check_binary_files.outputs.rust_files_changed }}
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-        with:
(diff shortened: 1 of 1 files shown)


Malax (changes requested):
This check is quite brittle - for example, people might [change files that are included to the code via macros](https://doc.rust-lang.org/std/macro.include.html), which will not be detected as a "code change" here.

I think re-publishing is an edge case, so forcing maintainers to remove a tag from the VCS by hand before re-publishing is a better solution since its more explicit and doesn't rely on heuristics alone.

edmorley:
@colincasey The PR has had commits since the last review. If it's ready for re-review, don't forget to re-request review using GitHub's UI :-)

edmorley:
This was superseded by #95.

PR #55, opened 2023-06-29
Remove setup-tools step from package workflow

### LINKED ISSUE #51: `setup-tools` is unused in the package buildpack workflow
The `_buildpacks-release-package.yml` workflow currently installs yj/crane/... here:
https://github.com/heroku/languages-github-actions/blob/b17eaa3156924ccb4fc6d38d2edd8ad7ac5617a5/.github/workflows/_buildpacks-release-package.yml#L50-L51

However, the package step doesn't use any of those tools. (Nothing in this repo uses yj, and only the update-builder and publish to CNB registry workflows use crane.)

colincasey:
The `setup-tools` step was previously used to install `yj` and extract metadata from the buildpack during the `package` workflow but this is no longer required since that metadata is now extracted as part of the `detect` workflow.

@fixes #51

### DIFF
--- a/.github/workflows/_buildpacks-release-package.yml
+++ b/.github/workflows/_buildpacks-release-package.yml
@@ -47,9 +47,6 @@ jobs:
       - name: Install Pack CLI
         uses: buildpacks/github-actions/setup-pack@v5.2.0
 
-      - name: Install yj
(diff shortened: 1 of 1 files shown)


edmorley (approved):
Thank you!

Issue #49, opened 2023-06-29
Components of meta-buildpacks are compiled multiple times which could result in non-determinism


edmorley:
Currently an initial detect job is run, which then fans out to multiple other jobs - one job for each buildpack.

In these other jobs, each individual buildpack (or meta-buildpack) is compiled (using `cargo libcnb package`) and then packaged (using `pack buildpack package`).

This means that component buildpacks (such as the Node.js engine buildpack, or the JVM buildpack) are compiled multiple times on different GitHub Actions runners, since they are compiled not only as standalone buildpacks, but again when their parent meta-buildpacks are built (since a meta-buildpack bundles its own dependencies).

If the result of the Cargo build happened to be non-deterministic, this could mean the contents of the layer differed between the standalone and meta-buildpack releases.

For example the contents of `heroku/buildpack-nodejs-engine` could differ from the engine subset of the contents of `heroku/buildpack-nodejs`. (I've checked now, and it seems for those two specific images, the builds for 1.1.0 happen to be deterministic, but there's no guarantee that would always be the case.)

This may cause issues in a few ways:
- Our builder images reference both meta-buildpacks (eg `heroku/nodejs`) and component buildpacks (eg `heroku/nodejs-engine` in Ruby's definition). How does `pack builder create` handle two buildpacks with the same ID and version having differing contents, when creating the builder? Does one of the contents just win over the other? Does the command error?
- Users could get different buildpack contents just depending on whether they wrote their own sub-component order in say `project.toml` vs using the meta-buildpack identifier

The way to avoid this would be to move the `cargo libcnb package` step earlier in the workflow, before the job matrix is created. 

This would:
- solve the non-determinism issue above
- save compiling similar things multiple times / reduce the number of CI runners consumed by each release (relevant if there is a GitHub Actions queuing backlog)
- also help fix #48

As an added bonus, the "detect" and "package" jobs could be combined into one, saving another CI runner.

cc @colincasey @Malax 

GUS-W-13738739.

edmorley:
Fixed by #91.

PR #57, opened 2023-06-29
Use unique cache keys in packaging workflow

### LINKED ISSUE #48: Rust caching of CNB compiles isn't working properly due to non-unique keys
The Rust cache for the buildpack packing step (that runs `cargo libcnb package`) currently uses a cache key of eg:
`v0-rust-release-package-9e858e73-0318b52e`

(see https://github.com/heroku/buildpacks-nodejs/actions/runs/5402321547/jobs/9813445930#step:29:11)

This means all buildpack package jobs share the same cache, hence messages like 

```
Failed to save: Unable to reserve cache with key v0-rust-release-package-9e858e73-0318b52e, another job may be creating this cache. More details: Cache already exists. Scope: refs/heads/main, Key: v0-rust-release-package-9e858e73-0318b52e, Version: f790786e8aef2abffd5cbe690195fe42930af1dfb65aca0111c17e682b11c013
```

(see https://github.com/heroku/buildpacks-nodejs/actions/runs/5402321547/jobs/9813445930#step:29:35)

Whilst at first glance it seems fine for all release compiles within the same repo to share the same cache, this is not the case because each buildpack is compiled separately (rather than the whole workspace being compiled as one), so the cache will only end up with a subset of the crates being compiled and cached.

One way to fix this would be to adjust the cache key settings here:
https://github.com/Swatinem/rust-cache#rust-cache-action

However, it may be best to switch to compiling the whole workspace as one to resolve other issues (see #49).

GUS-W-13738777.

colincasey:
Adding an extra cache key specifier to the rust caching step so that releases that need to package multiple buildpacks don't all try to use the same cache.

Fixes #48 

GUS-W-13738777.

### DIFF
--- a/.github/workflows/_buildpacks-release-package.yml
+++ b/.github/workflows/_buildpacks-release-package.yml
@@ -46,6 +46,8 @@ jobs:
 
       - name: Rust cache
         uses: Swatinem/rust-cache@v2
+        with:
+          key: ${{ inputs.buildpack_id }}-v${{ inputs.buildpack_version }}
 
       - name: Install Pack CLI
         uses: buildpacks/github-actions/setup-pack@v5.2.0


.github/workflows/_buildpacks-release-package.yml
```diff
@@ -46,6 +46,8 @@ jobs:
 
       - name: Rust cache
         uses: Swatinem/rust-cache@v2
+        with:
+          key: ${{ inputs.buildpack_id }}-v${{ inputs.buildpack_version }}
```
edmorley, line 50: I don't think we want the version here, since:
- the version will change for each release, meaning the cache is never reused from one release to the next
- the version isn't what affects the contents of the build, things like the `Cargo.lock` do, and the Rust caching plugin already uses the SHA256 of that file

Just the ID on it's own should be good I think.

edmorley (changes requested):

edmorley:
Also worth noting: This PR won't be needed when #49 is fixed -- it might be best to just focus on #49?

PR #75, opened 2023-07-20
Improved release process

### LINKED ISSUE #13: Publishing does not block when version already exists

colincasey:
This release process is less automated but is required for internally referenced actions which need their versions to match the published binary when a release is produced.

Fixes #13, [GUS-W-13658882](https://gus.lightning.force.com/a07EE00001Ul6LdYAJ)


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,86 @@
+name: Prepare Shared GitHub Actions for Release
+
+on:
+  push:
```
edmorley, line 4: Does the `push` event need to be removed here?

```suggestion
  push:
```

PR #69, opened 2023-07-19
Update to request-add-entry action v5.3.0

edmorley:
Dependabot can currently only update standard actions, not those that use a Docker Hub reference - so this Docker Hub reference has to be updated manually.

The upstream `request-add-entry` action has a number of fixes we should pick up:
https://github.com/buildpacks/github-actions/compare/v5.0.1...v5.3.0

Most notably:
- https://github.com/buildpacks/github-actions/pull/198
- https://github.com/buildpacks/github-actions/pull/214

Fixes #68.
GUS-W-13788681.

PR #84, opened 2023-07-24
Downgrade request-add-entry action to v5.2.0

Malax:
The image tag `5.3.0` cannot be found currently. Unsure if we updated without checking it actually exists or if the tag was removed after the fact. In any case, this PR downgrades to the latest version that is available on ghcr.io to unblock users of the actions.

Ref: https://github.com/heroku/languages-github-actions/pull/69

### DIFF
--- a/.github/workflows/_buildpacks-release-publish-cnb-registry.yml
+++ b/.github/workflows/_buildpacks-release-publish-cnb-registry.yml
@@ -42,7 +42,7 @@ jobs:
 
       - name: Register the new version with the CNB Buildpack Registry
         if: steps.check.outputs.published_to_cnb_registry == 'false'
-        uses: docker://ghcr.io/buildpacks/actions/registry/request-add-entry:5.3.0
+        uses: docker://ghcr.io/buildpacks/actions/registry/request-add-entry:5.2.0
         with:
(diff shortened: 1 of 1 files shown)


edmorley (approved):
I updated to 5.3.0 based on the release existing here:
https://github.com/buildpacks/github-actions/releases

Since (a) there was no direct link to the GCHR repo for me to check the tag, and (b) I presumed that project actually had a reliable release process.

But as you say, the 5.3.0 tag doesn't exist:
https://github.com/orgs/buildpacks/packages/container/package/actions%2Fregistry%2Frequest-add-entry

I would guess perhaps a human has to remember to manually build/push the image, sigh.

So one option would be for us to just switch to the latest tag for this repo, given Dependabot won't update it, that might actually reduce the burden? (Given this request-entry GitHub Action is fairly constrained in what it does, the potential for breakage seems like it would be acceptable?)

edmorley:
The latest tag is actually what is documented here:
https://github.com/buildpacks/github-actions#request-add-entry-action

Malax:
> So another option would be for us to just switch to the latest tag for this repo

Which also doesn't exist. 😭

edmorley:
Of course it doesn't 😆

PR #86, opened 2023-07-24
Clean up workflow permissions

### LINKED ISSUE #53: Are `GITHUB_TOKEN` write permissions needed for workflows that use the GitHub App?
In some workflows, permissions are set to `write`, eg:
https://github.com/heroku/languages-github-actions/blob/b17eaa3156924ccb4fc6d38d2edd8ad7ac5617a5/.github/workflows/_buildpacks-release-update-builder.yml#L17C1-L19

However, for the interactions with GitHub the token from the GitHub App is used:
https://github.com/heroku/languages-github-actions/blob/b17eaa3156924ccb4fc6d38d2edd8ad7ac5617a5/.github/workflows/_buildpacks-release-update-builder.yml#L56-L61
https://github.com/heroku/languages-github-actions/blob/b17eaa3156924ccb4fc6d38d2edd8ad7ac5617a5/.github/workflows/_buildpacks-release-update-builder.yml#L75-L77

...so it seems the `GITHUB_TOKEN` is not used for writes after all, and doesn't need the additional permissions?

colincasey:
Removed all unnecessary permission declarations. The GitHub App contains the required permissions to perform these actions.

Fixes #53, [W-13738819](https://gus.lightning.force.com/a07EE00001Vrz0XYAR)

### DIFF
--- a/.github/workflows/_buildpacks-release-detect.yml
+++ b/.github/workflows/_buildpacks-release-detect.yml
@@ -8,9 +8,6 @@ on:
         description: The list of {buildpack_id, buildpack_version, buildpack_dir, buildpack_artifact_prefix, docker_repository} values for a buildpack formatted as a JSON array
         value: ${{ jobs.release-detect.outputs.buildpacks }}
 
-permissions:
-  contents: read
-
 jobs:
   release-detect:
     name: Executing
--- a/.github/workflows/_buildpacks-prepare-release.yml
+++ b/.github/workflows/_buildpacks-prepare-release.yml
@@ -18,10 +18,6 @@ on:
         description: Private key of GitHub application (Linguist)
         required: true
 
-permissions:
-  contents: write
-  pull-requests: write
-
 jobs:
   prepare-release:
     name: Prepare Release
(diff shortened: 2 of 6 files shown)


.github/workflows/_buildpacks-release-detect.yml
```diff
@@ -8,9 +8,6 @@ on:
         description: The list of {buildpack_id, buildpack_version, buildpack_dir, buildpack_artifact_prefix, docker_repository} values for a buildpack formatted as a JSON array
         value: ${{ jobs.release-detect.outputs.buildpacks }}
 
-permissions:
-  contents: read
```
edmorley, line 12: I presume the `actions/checkout` action doesn't need `read` permissions since the repos are public?

I'm pretty sure removing this should work - worst case we can add it back later :-)

edmorley (approved):

.github/workflows/_buildpacks-release-detect.yml
colincasey: from my testing it didn't appear to be required but there may be subtle differences between my personal org i used for testing and the heroku org that might not immediately surface (e.g.; IP whitelisting). 

one other thing i found was that permissions set in the shared workflow don't actually matter except in that they sometimes help display a more informative error message (i.e.; something like "workflow X requested read permission but it was not given"). 

only the primary workflow that calls the shared workflows controls the permission set (which makes sense) so this change shouldn't actually affect anything except that error message.

PR #91, opened 2023-07-26
Fix rust caching and duplicate buildpack compilation

### LINKED ISSUE #48: Rust caching of CNB compiles isn't working properly due to non-unique keys

colincasey:
This is a restructuring of the whole release process to eliminate the issues outlines in #48 (buildpacks compiled multiple times could result in non-determinism) and #49 (rust caching isn't working due to non-unique keys).

Many of the overall steps are still the same but the `detect` job has been replaced by a `compile` job which front loads all the rust and buildpack compilation and cached the compiled buildpack folder for later use. Compilation was previously handled in the `package` job which is no longer needed.

To support this new flow, a `buildpack_output_dir` value had to be added to the generated buildpack matrix. This also allowed some simplification in terms of the assets cached since the OCI image and CNB file can be generated directly from the compiled `buildpack_output_dir` we don't need to store those values in the cache anymore.

Fixes #48 ([W-13738777](https://gus.lightning.force.com/a07EE00001Vs1EqYAJ)), #49 ([W-13738739](https://gus.lightning.force.com/a07EE00001VrgrNYAR))

Note - I discovered I was mistaken about how strategies and shared workflows interact.  Turns out you can use them. The GitHub Actions docs make it sound like you can't and I must have messed up when I initially tried this.  Using this new shared workflow should simplify the calling `release.yml` files in our buildpack repos 😄 

We'll also need to be careful with changes to buildpack output locations like https://github.com/heroku/libcnb.rs/pull/583 and ensure these still line up with expectations in our release process.

### DIFF
--- a/.github/workflows/_buildpacks-release.yml
+++ b/.github/workflows/_buildpacks-release.yml
@@ -1,0 +1,245 @@
+name: _buildpacks-release
+
+on:
+  workflow_call:
+    inputs:
+      app_id:
+        description: Application ID of GitHub application (Linguist)
+        type: string
+        required: true
+      dry_run:
+        required: false
+        type: boolean
+        default: false
+        description: Flag used for testing purposes that prevents actual changes to the heroku/builder repo
+      ip_allowlisted_image:
+        description: The image to use to run jobs that require IP allowlist privileges
+        type: string
+        required: false
+        default: pub-hk-ubuntu-22.04-small
+    secrets:
+      app_private_key:
+        description: Private key of GitHub application (Linguist)
+        required: true
+      cnb_registry_token:
+        required: true
+        description: The token to login to the CNB registry with
+      docker_hub_user:
+        required: true
+        description: The username to login to Docker Hub with
+      docker_hub_token:
+        required: true
+        description: The token to login to Docker Hub with
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  compile:
+    name: Compile Buildpacks
+    runs-on: ubuntu-latest
+    outputs:
+      buildpacks: ${{ steps.generate-buildpack-matrix.outputs.buildpacks }}
+      version: ${{ steps.generate-buildpack-matrix.outputs.version }}
+      changelog: ${{ steps.generate-changelog.outputs.changelog }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: true
+
+      - name: Install musl-tools
+        run: sudo apt-get install musl-tools --no-install-recommends
+
+      - name: Update Rust toolchain
+        run: rustup update
+
+      - name: Install Rust linux-musl target
+        run: rustup target add x86_64-unknown-linux-musl
+
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2.5.1
+
+      - name: Install libcnb-cargo
+        run: cargo install libcnb-cargo
+
+      - name: Package buildpacks
+        id: libcnb-package
+        run: cargo libcnb package --release
+
+      - name: Generate buildpack matrix
+        id: generate-buildpack-matrix
+        uses: heroku/languages-github-actions/.github/actions/generate-buildpack-matrix@main
+
+      - name: Generate changelog
+        id: generate-changelog
+        uses: heroku/languages-github-actions/.github/actions/generate-changelog@main
+        with:
+          version: ${{ steps.generate-buildpack-matrix.outputs.version }}
+
+      - name: Cache buildpacks
+        uses: actions/cache/save@v3
+        with:
+          key: ${{ github.run_id }}-compiled-buildpacks
+          path: target/buildpack
+
+  publish-docker:
+    name: Publish → Docker - ${{ matrix.buildpack_id }}
+    needs: [compile]
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include: ${{ fromJSON(needs.compile.outputs.buildpacks) }}
+    steps:
+      - name: Restore buildpacks
+        uses: actions/cache/restore@v3
+        with:
+          key: ${{ github.run_id }}-compiled-buildpacks
+          path: target/buildpack
+
+      - name: Install Pack CLI
+        uses: buildpacks/github-actions/setup-pack@v5.3.0
+
+      - name: Create Docker Image
+        run: pack buildpack package ${{ matrix.buildpack_id }} --config ${{ matrix.buildpack_output_dir }}/package.toml -v
+
+      - name: Login to Docker Hub
+        if: inputs.dry_run == false
+        uses: docker/login-action@v2.2.0
+        with:
+          registry: docker.io
+          username: ${{ secrets.docker_hub_user }}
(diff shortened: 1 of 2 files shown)


.github/workflows/_buildpacks-release.yml
```diff
@@ -0,0 +1,245 @@
...
+        default: false
+        description: Flag used for testing purposes that prevents actual changes to the heroku/builder repo
+      ip_allowlisted_image:
+        description: The image to use to run jobs that require IP allowlist privileges
+        type: string
+        required: false
+        default: pub-hk-ubuntu-22.04-small
+    secrets:
+      app_private_key:
+        description: Private key of GitHub application (Linguist)
```
edmorley, line 22:
```suggestion
        description: Private key of GitHub application (e.g. the Linguist App)
```

.github/workflows/_buildpacks-release.yml
```diff
@@ -0,0 +1,245 @@
...
+        required: true
+      cnb_registry_token:
+        required: true
+        description: The token to login to the CNB registry with
```
edmorley, line 26: The CNB registry itself doesn't  have its own auth system, the token is that of a GitHub account:
https://github.com/buildpacks/github-actions#request-add-entry-action

```suggestion
        description: The token of the GitHub user used to interact with the CNB registry
```

.github/workflows/_buildpacks-release.yml
```diff
@@ -0,0 +1,245 @@
...
+        id: generate-changelog
+        uses: heroku/languages-github-actions/.github/actions/generate-changelog@main
+        with:
+          version: ${{ steps.generate-buildpack-matrix.outputs.version }}
+
+      - name: Cache buildpacks
+        uses: actions/cache/save@v3
+        with:
+          key: ${{ github.run_id }}-compiled-buildpacks
+          path: target/buildpack
```
edmorley, line 84: I wonder if the `generate_buildpack_matrix` command should output this value, and then it could be accessed via eg:
`${{ steps.generate-buildpack-matrix.outputs.output_dir }}`
...or similar?

That way all the Rust/libcnb related paths are at least in one place - given there is already the path `target/` hardcoded in `src/commands/generate_buildpack_matrix/command.rs` here:
https://github.com/heroku/languages-github-actions/pull/91/files#diff-7abed701849efef91f6445231b0efc6baa56c948c52a602b1028e4aeb554e4ccR81

Longer term, maybe `cargo libcnb package` should be made to accept an `--output-dir` path, which would mean the workflows in this repo would be in full control of the paths in question, so don't need to hardcode paths?

.github/workflows/_buildpacks-release.yml
```diff
@@ -0,0 +1,245 @@
...
+        uses: actions/checkout@v3
+        with:
+          repository: heroku/builder
+          path: ./builder
+
+      - name: Install crane
+        uses: buildpacks/github-actions/setup-tools@v5.3.0
+
+      - name: Update Builder
+        if: inputs.dry_run == false
```
edmorley, line 214: Should this `if: inputs.dry_run == false` conditional be moved to the "create pull request" step?

Whilst the "create pull request" step should skip itself if there are no changed files (so the current implementation won't write changes), moving the conditional would:
- make the intent here a bit easier to follow
- mean the update-builder command gets run/tested in dry run mode (whereas at the moment it doesn't run, so isn't tested)

edmorley (commented):
Thank you for doing this! I love the new combined shared workflow idea too - as you say, it should reduce the boilerplate in each buildpack's repo :-)

I presume the rollout plan for this is to merge this PR, switch the repos over to using the new workflow, then remove the old separated workflows afterwards? (If so, that sounds good to me - I was just initially wondering why the old workflows weren't deleted in this PR.)

colincasey:
@edmorley re: rollout plan - you are correct.  merge this PR, switch the repos over to using the new workflow, then remove the old separated workflows afterwards.

.github/workflows/_buildpacks-release.yml
```diff
@@ -0,0 +1,245 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      buildpacks: ${{ steps.generate-buildpack-matrix.outputs.buildpacks }}
+      version: ${{ steps.generate-buildpack-matrix.outputs.version }}
+      changelog: ${{ steps.generate-changelog.outputs.changelog }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: true
```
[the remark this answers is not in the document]
colincasey: yes, we added this during testing of the JVM repo.  we could switch this to an optional input if we wanted to optimize but keeping it enabled as a default seems to work well currently.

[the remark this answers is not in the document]
edmorley: My point was that I thought this was only needed for tests. ie: It should be possible to remove it. Is that not the case? If so, what fails?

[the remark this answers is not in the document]
colincasey: not just for tests, no.  the [sbt-extras scripts](https://github.com/heroku/buildpacks-jvm/tree/main/buildpacks/sbt) are included as git submodules also.  when they are not present the `heroku/sbt` buildpack fails packaging when it hits this:

```
buildpacks/sbt/src/layers/sbt_extras.rs:44:13
   |
44 |             include_bytes!("../../sbt-extras/sbt"),
```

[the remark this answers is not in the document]
edmorley: Ah yes, thank you

.github/workflows/_buildpacks-release.yml
colincasey: i like the longer term idea that we control that output path through the whole process.  in that case i think we'd define it as an optional input to the workflow.  

for now, i'm not happy to have hardcoded aspects but i think this is fine for now.  i'll create a new work item that outlines what changes to `libcnb-cargo` and this repo would be needed to coordinate this.

edmorley (approved):

PR #95, opened 2023-07-31
Improved release process

### LINKED ISSUE #13: Publishing does not block when version already exists

colincasey:
This release process supports the following for projects using shared workflows:

- Development and testing of automation changes from the `main` branch 
- Stable, deterministic builds from `v{major}.{minor}.{patch}` refs

For development releases:

- `.github/actions/install-languages-cli/action.yml` must contain an empty value as the default for `download_url` 
> This will trigger the CLI application to be downloaded and installed via Cargo in shared workflows so that the latest changes to Rust code will be used
- Internal actions within shared workflows are locked to the `main` ref so that latest changes to automation will be used
  
For stable releases:
- The previous release version will be bumped according to which option was chosen (`major`, `minor`, or `patch`) when triggering the release workflow
- The Cargo version will be set to this new version
- Internal actions within shared workflows will be updated to use the new version as their ref
- `.github/actions/install-languages-cli/action.yml` will be populated with the location of the CLI asset uploaded to the release as the default value for `download_url` 
> This will trigger the CLI application to be downloaded and installed from a versioned release in shared workflows
- These changes will be committed to a release tag and a new GitHub Release will be created from that

Fixes #13, GUS-W-13658882


.github/workflows/_buildpacks-prepare-release.yml
```diff
@@ -23,30 +33,40 @@ on:
...
-        uses: heroku/languages-github-actions/.github/actions/prepare-release@main
-        with:
-          bump: ${{ inputs.bump }}
-          declarations_starting_version: ${{ inputs.declarations_starting_version }}
+        run: |
+          ARGS=("--bump" "${{ inputs.bump }}" "--repository-url" "https://github.com/${{ github.repository }}")
+          if [ -n "${{ inputs.declarations_starting_version }}" ]; then
+            ARGS+=("--declarations-starting-version" "${{ inputs.declarations_starting_version }}")
+          fi
+          actions prepare-release "${ARGS[@]}"
```
edmorley, line 65: Would something like this work, to avoid the need to generate the args list? 

This makes use of the `format()` function:
https://docs.github.com/en/actions/learn-github-actions/expressions#format

...plus their ternery-like feature:
https://docs.github.com/en/actions/learn-github-actions/expressions#example

```suggestion
        run: actions prepare-release --bump "${{ inputs.bump }}" --repository-url "https://github.com/${{ github.repository }}" ${{ inputs.declarations_starting_version && format('--declarations-starting-version {0}', inputs.declarations_starting_version) }}
```

.github/workflows/_buildpacks-prepare-release.yml
```diff
@@ -23,30 +33,40 @@ on:
...
+        uses: heroku/use-app-token-action@main
+        id: generate-token
+        with:
+          app_id: ${{ inputs.app_id }}
+          private_key: ${{ secrets.app_private_key }}
+
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          token: ${{ steps.generate-token.outputs.app_token }}
```
edmorley, line 48: The repo is public, so I would have thought the token isn't needed? Or is this as a way to test the token works early, rather than only find out right at the end when the "open a PR" step happens? (Or in the case of dry-run mode, the open a PR step doesn't happen at all, so wouldn't be tested.)

edmorley (commented):
This looks great so far! 

I've left some initial comments and can have another looks tomorrow :-)

.github/workflows/release.yml
```diff
@@ -0,0 +1,127 @@
...
+      - name: Update actions/install-languages-cli/action.yml
+        uses: mikefarah/yq@v4.34.2
+        with:
+          cmd: |
+            yq -i '
+              .inputs.download_url.default = "https://github.com/${{ github.repository }}/releases/download/v${{ steps.metadata.outputs.version }}/${{ steps.metadata.outputs.name }}.tar.gz"
+            ' .github/actions/install-languages-cli/action.yml
+
+      - name: Update workflows/_buildpacks-prepare-release.yml
+        uses: mikefarah/yq@v4.34.2
```
edmorley, line 88: The GitHub Actions runners already have `yq` installed:
https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md?plain=1#L104

Could we run it directly, saving the need to rely upon the `mikefarah/yq` action? (This would reduce the supply chain surface area improving security, and mean one less thing that results in Dependabot PR noise.)

.github/actions/install-languages-cli/action.yml
```diff
@@ -0,0 +1,34 @@
...
+      shell: bash
+      if: inputs.download_url == ''
+      run: cargo install --git https://github.com/heroku/languages-github-actions.git
+
+    - name: Download actions binary
+      shell: bash
+      if: inputs.download_url != ''
+      run: |
+        mkdir -p ${{ runner.tool_cache }}/languages-github-actions
+        curl ${{ inputs.download_url }} --output ${{ runner.temp }}/actions.tar.gz -L --retry 5 --retry-all-errors
```
edmorley, line 32: (The `--fail` ensures curl exits non-zero rather than the script trying to run `tar` anyway, and the timeout options prevent hangs)

```suggestion
        curl ${{ inputs.download_url }} --output ${{ runner.temp }}/actions.tar.gz -L --fail --retry 5 --retry-all-errors --connect-timeout 10 --max-time 60
```

(Also I've checked that it's ok to use `` since it was only added in 7.71.0, but it's fine since the GHA 22.04 runner [has v7.81.0-1ubuntu1.13](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#installed-apt-packages))

.github/actions/install-languages-cli/action.yml
```diff
@@ -0,0 +1,34 @@
+name: Install Languages CLI
+description: Dowloads the Languages CLI from a known release or using Cargo
```
edmorley, line 2:
```suggestion
description: Downloads the Languages CLI from a known release or installs using Cargo
```

.github/workflows/_buildpacks-release.yml
```diff
@@ -212,29 +224,25 @@ jobs:
...
+        uses: heroku/languages-github-actions/.github/actions/install-languages-cli@main
 
-      - name: Get token for GitHub application (Linguist)
-        uses: heroku/use-app-token-action@main
-        id: generate-token
-        with:
-          app_id: ${{ inputs.app_id }}
-          private_key: ${{ secrets.app_private_key }}
+      - name: Update Builder
+        if: inputs.dry_run == false
```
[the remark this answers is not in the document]
colincasey: it was added back in intentionally.  i realized during testing that the reason i had it set on this step was that the update process requires a published image to exist in order to calculate its digest with `crane` or it fails.  adding the dry run check here means that no modifications will occur and the following PR-related steps will do no work. i'd like to revisit at a later stage though.

.github/workflows/_buildpacks-prepare-release.yml
colincasey: hmm, would have to try that to see if it works.

.github/workflows/_buildpacks-prepare-release.yml
colincasey: oh, so this is actually related to how you can get signed commits for your GH App.  it's possible to do a bunch of the git setup manually but, if provided a token, the checkout command will automatically configure the git client with the required credentials to use for later commits that can be attributed to the bot.

.github/workflows/release.yml
```diff
@@ -0,0 +1,127 @@
+name: Release Actions
+
+env:
+  BOT_USER: heroku-linguist[bot]
+  BOT_EMAIL: 136119646+heroku-linguist[bot]@users.noreply.github.com
```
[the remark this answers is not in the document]
colincasey: this is also related to getting signed bot commits for the GH App.  if you don't set this then the default committer will be the user who triggered the workflow. 

i like the idea of having it automatically set by https://github.com/heroku/use-app-token-action though. i believe they would have all the information to dynamically fetch this based on the APP_ID we're using.

.github/workflows/_buildpacks-release.yml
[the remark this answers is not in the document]
edmorley: Ah that makes sense. Could you add a brief comment mentioning why the `dry-run` is on that step?

.github/workflows/release.yml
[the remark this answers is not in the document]
colincasey: actually, what might be better is if these were set as organization variables.  that way [heroku/use-app-token-action](https://github.com/heroku/use-app-token-action) would not need to be modified to make extra API calls to retrieve information that is specific to our use case.

.github/workflows/_buildpacks-prepare-release.yml
```diff
@@ -23,30 +33,41 @@ on:
...
+        uses: heroku/use-app-token-action@main
+        id: generate-token
+        with:
+          app_id: ${{ inputs.app_id }}
+          private_key: ${{ secrets.app_private_key }}
+
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          # The token is used to set the local git config which is used to run authenticated git commands
```
edmorley, line 48: I think this new comment is missing the key part - which is answering "why do I need a token for a public repo?". Instead the comment states something that IMO can already be seen/figured out by reading the code (that a token is being set for git auth).

In https://github.com/heroku/languages-github-actions/pull/95#discussion_r1283270681, you mention using the token ensures that git signing is set up as expected - so I think the comment needs to say something roughly like:

> "We set the token here even though it's checking out a public repo, because it ensures XYZ is set up as expected for when we commit changes later." 

...where XYZ is something about git signing.

.github/workflows/_buildpacks-release.yml
```diff
@@ -212,29 +224,29 @@ jobs:
...
 
-      - name: Get token for GitHub application (Linguist)
-        uses: heroku/use-app-token-action@main
-        id: generate-token
-        with:
-          app_id: ${{ inputs.app_id }}
-          private_key: ${{ secrets.app_private_key }}
+      - name: Update Builder
+        # The dry run check is performed here because the update process requires a published to exist in oder to calculate a digest with `crane`
+        # or it fails. Adding the check here means no files will be modified and no PR will be created for the PR-related steps that follow.
```
edmorley, line 244: (there were a few missing words/typos, and I have shortened slightly)

```suggestion
        # The dry run check is performed here because the update process requires a published
        # image to exist in order to calculate a digest with `crane`. Adding the check here
        # means no files will be modified and so no PR will be created later.
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,123 @@
...
+      - name: Get previous release version
+        id: previous-version
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.app_token }}
+        run: echo "value=$((gh release view --json tagName --jq '.tagName' 2> /dev/null || echo "0.0.0") | sed 's/^v//')" >> $GITHUB_OUTPUT
+
+      - name: Bump version
+        run: |
+          cargo bump ${{ steps.previous-version.outputs.value }}
+          cargo bump ${{ inputs.bump }}
```
edmorley, line 61: The `|| echo "0.0.0"` fallback I think adds a failure mode that would result in the older tags being overwritten if the `gh release` command were to fail for say connectivity reasons?

eg imagine:
1. `gh release` fails, even though there is an existing release (eg `1.2.0`), due to network issues or GitHub having a bad day
2. Instead of the workflow aborting/failing, the `|| echo "0.0.0"` fallback means `value` is set to `0.0.0`
3. As such the first bump command ends up being `cargo bump 0.0.0`, meaning the bump following it (eg `cargo bump minor`) would then turn the version into `0.1.0` - even though `0.1.0` is actually older than the current `1.2.0` release.

To avoid this, either:
1. The implementation needs to tell the difference between "there are no existing releases" vs "the gh release command failed for a different reason". (Which is probably more hassle than it's worth to implement)
2. Or, we just remove the `|| echo "0.0.0"` and accept the fact that the implementation here assumes there is an existing release.

Given that we know an existing release exists for this repo, I think (2) might be the easier option?

Also, I think the `2> /dev/null` might (a) be unnecessary (given that pipes and redirects only redirect stdout by default), and (b) might be best removed so if any error occurs it appears in the GitHub Actions log.

.github/workflows/release.yml
```diff
@@ -77,51 +77,47 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ steps.generate-token.outputs.app_token }}
         with:
-          target_commitish: v${{ steps.metadata.outputs.version }}
+          target_commitish: ${{ steps.release-tag.outputs.sha }}
           tag_name: v${{ steps.metadata.outputs.version }}
           files: ${{ steps.package-binary.outputs.path }}
           fail_on_unmatched_files: true
-          body: "Full Changelog: https://github.com/${{ github.repository }}/compare/v${{ steps.previous-version.outputs.value }}..v${{ steps.metadata.outputs.version }}"
+          body: "Full Changelog: https://github.com/${{ github.repository }}/compare/${{ steps.previous-version.outputs.value == '0.0.0' && 'main' || format('v{0}', steps.previous-version.outputs.value) }}..v${{ steps.metadata.outputs.version }}"
```
edmorley, line 123: If the `0.0.0` fallback is removed higher up in the workflow (xref the other thread on this review), then this fallback will need to be removed too.

.github/workflows/build-test-release.yml
```diff
@@ -1,126 +0,0 @@
...
-          node-version: lts/*
-          cache: npm
-      - name: Install Node dependencies
-        run: npm ci
-      - name: Rebuild bootstrap script
-        run: npm run build
-      - name: Check bootstrap script for uncommitted changes
-        run: if [ -n "$(git status --porcelain)" ]; then exit 1; fi
-
-  release:
```
edmorley, line 63: With the removal of the old release workflow (plus once #97 is merged), I believe all of the following can be deleted:

```
.github/actions/<everything except the new install cli action>
.github/bootstrap/*
package-lock.json
package.json
tsconfig.json
```

Plus the `node_modules` entry in `.gitignore`.

.github/workflows/release.yml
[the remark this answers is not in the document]
edmorley: Is the plan to switch to org variables now (and remove these hardcoded lines), or later? (I'm fine with later, it just wasn't clear since this conversation was marked as resolved without context)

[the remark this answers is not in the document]
colincasey: sorry, i'll get that sorted out later since it'll require filing a request against another team

.github/workflows/release.yml
colincasey: yes, i was trying to accommodate the "no previous release" edge case but we know there is one already so i'll drop this extra logic.

.github/workflows/build-test-release.yml
colincasey: done + updated the README to reflect the new setup

README.md
```diff
@@ -4,22 +4,25 @@ A set of custom GitHub Actions and reusable workflow used by the Languages Team.
...
-* opening a PR against the repository with the modified files
+- bumping the fixed version
+- updating changelogs
+- generating an aggregate changelog from all the changelogs
+- opening a PR against the repository with the modified files
+
+You can pin to:
+- the [latest release](/releases/latest) version with `latest`
+- a [specific release](/releases) version with `@v{major}.{minor}.{patch}`
+- the development version with `main`
```
edmorley, line 25: If I click the links when having opened the `README.md` directly (as opposed to viewing when looking at the GitHub overview UI), they currently 404. 

eg Visit https://github.com/heroku/languages-github-actions/blob/06ac54789d81ade15ace09a6e1fa18ae1860b659/README.md and click the links.

I think the only way to avoid that is having to include the full URL sadly.

Also, currently some of the examples are prefixed with `@` but some aren't - I think for clarity/consistency we should prefix them all with `@`.

```suggestion
- the [latest release](https://github.com/heroku/languages-github-actions/releases/latest) version with `@latest`
- a [specific release](https://github.com/heroku/languages-github-actions/releases) version with `@v{major}.{minor}.{patch}`
- the development version with `@main`
```

README.md
```diff
@@ -66,11 +70,20 @@ jobs:
...
+  * Creates a CNB archive file from the compiled buildpack and publishes it as a GitHub Release
+  * Creates an OCI image from the compiled buildpack and publishes it to Docker
+  * Retrieves the OCI image published to Docker and registers this with the CNB Registry
+* Once all buildpacks have been published, all the buildpack references found in [heroku/builder](https://github.com/heroku/builder)
+  for the given list of builders and opens a pull requests containing all the changes to be committed.
+
+You can pin to:
+- the [latest release](/releases/latest) version with `latest`
+- a [specific release](/releases) version with `@v{major}.{minor}.{patch}`
+- the development version with `main`
```
edmorley, line 86:
```suggestion
- the [latest release](https://github.com/heroku/languages-github-actions/releases/latest) version with `@latest`
- a [specific release](https://github.com/heroku/languages-github-actions/releases) version with `@v{major}.{minor}.{patch}`
- the development version with `@main`
```

README.md
```diff
@@ -113,125 +126,33 @@ jobs:
...
-  uses: heroku/languages-github-actions/.github/actions/prepare-release@main
+- name: Install Languages CLI
+  uses: heroku/languages-github-actions/.github/actions/install-languages-cli@latest
 ```
 
-You can also pin to a [specific release](/releases) version in the format `@v{major}.{minor}.{patch}`
+You can pin to:
+- the [latest release](/releases/latest) version with `latest`
+- a [specific release](/releases) version with `@v{major}.{minor}.{patch}`
+- the development version with `main`
```
edmorley, line 143:
```suggestion
- the [latest release](https://github.com/heroku/languages-github-actions/releases/latest) version with `@latest`
- a [specific release](https://github.com/heroku/languages-github-actions/releases) version with `@v{major}.{minor}.{patch}`
- the development version with `@main`
```

README.md
```diff
@@ -41,7 +44,7 @@ on:
 
 jobs:
   prepare-release:
-    uses: heroku/languages-github-actions/.github/workflows/_buildpacks-prepare-release.yml@main
+    uses: heroku/languages-github-actions/.github/workflows/_buildpacks-prepare-release.yml@latest
     with:
       app_id: ${{ vars.GH_APP_ID }}
```
edmorley, line 49: I presume this is `vars.GH_APP_ID` instead of `vars.LINGUIST_GH_APP_ID` (and similar for the secrets section) since you're wanting to keep the README generic, and not tied to our specific Linguist implementation choices? (I'm happy with either approach, so how it is at the moment is fine - I'm only mentioning in case this naming wasn't intentional)

README.md
```diff
@@ -66,11 +70,20 @@ jobs:
 
 ### Buildpacks - Release
 
-Prepares a buildpack release by:
-* bumping the fixed version
-* updating changelogs
-* generating an aggregate changelog from all the changelogs
-* opening a PR against the repository with the modified files
+Prepares one or more buildpacks release by:
```
edmorley, line 73: I think this sentence needs adjusting, since it mentions preparing a release, but this workflow is the "perform a release" one?

README.md
```diff
@@ -66,11 +70,20 @@ jobs:
...
+
+* Detects all the buildpacks in a repository and compiles them into Cloud Native Buildpacks
+* For each compiled buildpack:
+  * Creates a CNB archive file from the compiled buildpack and publishes it as a GitHub Release
+  * Creates an OCI image from the compiled buildpack and publishes it to Docker
```
edmorley, line 78: (Feel free to reword this suggestion further - it's still missing an explanation of `[metadata.release]` etc, though maybe that's good enough for now?)

```suggestion
  * Creates an OCI image from the compiled buildpack and publishes it to the Docker Hub repository specified in the buildpack's `buildpack.toml`
```

README.md
```diff
@@ -66,11 +70,20 @@ jobs:
...
+  * Retrieves the OCI image published to Docker and registers this with the CNB Registry
```
edmorley, line 79:
```suggestion
  * Retrieves the OCI image published to Docker Hub and registers this with the CNB Registry
```

README.md
```diff
@@ -66,11 +70,20 @@ jobs:
...
+* Once all buildpacks have been published, all the buildpack references found in [heroku/builder](https://github.com/heroku/builder)
```
edmorley, line 80:
```suggestion
* Once all buildpacks have been published, updates all the buildpack references found in [heroku/builder](https://github.com/heroku/builder)
```

README.md
```diff
@@ -113,125 +126,33 @@ jobs:
...
+### Install Languages CLI
 
-| Name        | Description                          |
-|-------------|--------------------------------------|
-| `changelog` | Markdown content listing the changes |
-
-### Prepare Release
-
-Bumps the version of each detected buildpack and adds an entry for any unreleased changes from the changelog.
+Downloads the Languages CLI from a known release or installs using Cargo
```
edmorley, line 131:
```suggestion
Downloads the Languages CLI from a known release or installs using Cargo.
```

README.md
```diff
@@ -52,11 +55,12 @@ jobs:
...
+| Name                            | Description                                                                                                                                                                                             | Required | Default                     |
+|---------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------|-----------------------------|
+| `app_id`                        | Application ID of GitHub application (e.g. the Linguist App)                                                                                                                                            | true     |                             |
+| `bump`                          | Bump which coordinate? (major, minor, patch)                                                                                                                                                            | true     |                             |
```
edmorley, line 61:
```suggestion
| `bump`                          | Which component of the version to increment (major, minor, or patch)                                                                                                                                                            | true     |                             |
```

.github/workflows/_buildpacks-prepare-release.yml
```diff
@@ -1,22 +1,32 @@
...
-      bump:
-        description: Bump which coordinate? (major, minor, patch)
-        required: true
-        type: string
       app_id:
         description: Application ID of GitHub application (e.g. the Linguist App)
         type: string
         required: true
+      bump:
+        description: Bump which coordinate? (major, minor, patch)
```
edmorley, line 16:
```suggestion
        description: Which component of the version to increment (major, minor, or patch)
```

edmorley (approved):
Amazing thank you! 😄

.github/workflows/release.yml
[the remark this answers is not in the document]
edmorley: I've filed #99 to track moving these to org variables :-)

README.md
colincasey: yeah, it was to keep the README generic

README.md
colincasey: done

README.md
colincasey: added information about the metadata key used

PR #111, opened 2023-08-09
Fix GitHub publish

### LINKED ISSUE #89: Duplicate deploy behavior unknown
I triggered a release (intentionally) before triggering a prepare (unintentionally) for the `heroku/ruby` CNB. 

- Release link: https://github.com/heroku/buildpacks-ruby/actions/runs/5659205505/job/15332369486

## Expected

I'm not sure. Maybe some warning or error that the component had already been released. I like the rubygems version policy where versions are immutable, so there's never a case where "this version" means something different. Possibly disallowing deployment of a buildpack if it's got stuff in the "unreleased" section.

## Actual

I don't actually know all of what happened. Ed pointed this out:

>  just got a GitHub release notification for Ruby CNB v2.0.0, which was already released - so I think the old git tag will have been moved from the previous git SHA to a newer git sha?
https://github.com/heroku/buildpacks-ruby/releases/tag/v2.0.0
> I don't think any Docker Hub and/or CNB registry changes were made (since those steps skip if there is already the same version published), however, it looks like the GitHub releases part of the CNB automation (which includes the Git tagging part too) doesn't skip if the release already exists:
https://github.com/heroku/buildpacks-ruby/actions/runs/5659205505/job/15332369486

colincasey:
Changes the GitHub release process from being performed using a strategy for each buildpack which would perform partial overwrites of the release details and assets to packaging all the `*.cnb` files using a bash loop + `jq` and creating a single release with all the assets attached.

With this modification, a guard check can now be added to check if the release exists before attempting to create the release.  The "Create GitHub Release" step will be skipped if the release already exists or we're running the actions in `dry_run=true` mode. This should prevent a GH release from being clobbered if the release workflow is accidentally triggered without running the prepare release workflow first.

Fixes #89, [W-13871126](https://gus.lightning.force.com/a07EE00001XRghXYAT)

### DIFF
--- a/.github/workflows/_buildpacks-release.yml
+++ b/.github/workflows/_buildpacks-release.yml
@@ -166,10 +166,7 @@ jobs:
         id: check
         env:
           GH_TOKEN: ${{ steps.generate-token.outputs.app_token }}
-        run: echo "published_to_github=$(gh release view v0.0.3 &> /dev/null && echo "true" || echo "false")" >> $GITHUB_OUTPUT
-
-      - name: Test
-        run: echo "${{ steps.check.outputs.published_to_github }}"
+        run: echo "published_to_github=$(gh release view v${{ needs.compile.outputs.version }} &> /dev/null && echo "true" || echo "false")" >> $GITHUB_OUTPUT
 
       - name: Create GitHub Release
         if: steps.check.outputs.published_to_github == 'false'
--- a/.github/workflows/_buildpacks-release.yml
+++ b/.github/workflows/_buildpacks-release.yml
@@ -149,10 +149,10 @@ jobs:
 
       - name: Generate CNB files
         run: |
-          for buildpack in $(echo '${{ needs.compile.outputs.buildpacks }}' | jq -c '.[]'); do
-            artifact_prefix=$(echo "${buildpack}" | jq -r '.buildpack_artifact_prefix')
-            output_dir=$(echo "${buildpack}" | jq -r '.buildpack_output_dir')
-            pack buildpack package ${artifact_prefix}.cnb --config ${output_dir}/package.toml --format file -v
+          for buildpack in $(jq -c '.[]' <<< '${{ needs.compile.outputs.buildpacks }}'); do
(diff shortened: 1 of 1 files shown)


.github/workflows/_buildpacks-release.yml
```diff
@@ -151,8 +147,13 @@ jobs:
       - name: Install Pack CLI
         uses: buildpacks/github-actions/setup-pack@v5.3.1
 
-      - name: Save CNB File
-        run: pack buildpack package ${{ matrix.buildpack_artifact_prefix }}.cnb --config ${{ matrix.buildpack_output_dir }}/package.toml --format file -v
+      - name: Save CNB Files
```
edmorley, line 150:
```suggestion
      - name: Generate CNB files
```

.github/workflows/_buildpacks-release.yml
```diff
@@ -151,8 +147,13 @@ jobs:
...
+        run: |
+          for buildpack in $(echo '${{ needs.compile.outputs.buildpacks }}' | jq -c '.[]'); do
+            artifact_prefix=$(echo "${buildpack}" | jq -r '.buildpack_artifact_prefix')
+            output_dir=$(echo "${buildpack}" | jq -r '.buildpack_output_dir')
+            pack buildpack package ${artifact_prefix}.cnb --config ${output_dir}/package.toml --format file -v
+          done
```
edmorley, line 156: (using bash [here string](https://linux.die.net/abs-guide/x15683.html) saves having to do the echo+pipe, the quoting changes are suggestions from shellcheck)

```suggestion
          for buildpack in $(jq -c '.[]' <<< '${{ needs.compile.outputs.buildpacks }}'); do
            artifact_prefix=$(jq -r '.buildpack_artifact_prefix' <<< "${buildpack}")
            output_dir=$(jq -r '.buildpack_output_dir' <<< "${buildpack}")
            pack buildpack package "${artifact_prefix}.cnb" --config "${output_dir}/package.toml" --format file --verbose
          done
```

edmorley (approved):
Awesome!

Issue #125, opened 2023-08-19
Upgrade to libcnb 0.14.0 and remove hardcoded directory paths


edmorley:
This repo is currently using libcnb 0.13.0.

In order to update to libcnb 0.14.0 some changes will be required, see:
https://github.com/heroku/languages-github-actions/pull/120#issuecomment-1684407167

GUS-W-13977418.

PR #135, opened 2023-08-29
Keep `libcnb.rs` tools in sync

colincasey:
Because the languages CLI tooling depends on `libcnb-package` to determine build information, this should be kept in sync with the version of `libcnb-cargo` installed in the `_buildpacks-release.yml` workflow so that we can be sure that behavior of the `libcnb.rs` tooling is consistent.

Related to #120 and #125

### DIFF
--- a/.github/workflows/_buildpacks-release.yml
+++ b/.github/workflows/_buildpacks-release.yml
@@ -70,10 +70,12 @@ jobs:
       - name: Rust cache
         uses: Swatinem/rust-cache@v2.6.2
 
+      # the version of `libcnb-cargo` installed here is kept in sync with the version of `libcnb-package`
+      # that the release automation CLI tooling depends on
       - name: Install libcnb-cargo
-        # TODO: Derive this version from the `libcnb-package` version in this repo's `Cargo.toml`,
-        # or at least add a CI check to ensure the two versions are in sync.
-        run: cargo install --locked libcnb-cargo@0.13.0
+        run: |
+          LIBCNB_PACKAGE_VERSION=$(yq -ptoml -oyaml '.package[] | select(.name == "libcnb-package") | .version' Cargo.lock)
+          cargo install --locked "libcnb-cargo@${LIBCNB_PACKAGE_VERSION}"
 
       - name: Install Languages CLI
         uses: heroku/languages-github-actions/.github/actions/install-languages-cli@main


Malax (approved):

.github/workflows/_buildpacks-release.yml
```diff
@@ -70,10 +70,12 @@ jobs:
       - name: Rust cache
         uses: Swatinem/rust-cache@v2.6.2
 
+      # the version of `libcnb-cargo` installed here is kept in sync with the version of `libcnb-package`
+      # that the release automation CLI tooling depends on
       - name: Install libcnb-cargo
-        # TODO: Derive this version from the `libcnb-package` version in this repo's `Cargo.toml`,
```
edmorley, line 74: The comment here can now be removed too :-)
https://github.com/heroku/languages-github-actions/blob/e9a028c29594d48faf9d9b0123b333c8bcd63cb4/Cargo.toml#L28

edmorley:
@colincasey This is ready to merge once the one review comment above is addressed :-)

.github/workflows/_buildpacks-release.yml
colincasey: removed

colincasey:
@edmorley @Malax this is going to need a re-review.  

during testing i realized there was a major flaw in this approach.  the workflow executing doesn't see the `Cargo.toml` from this repository, rather it sees the `Cargo.toml` from the calling repository.  i had to restructure the logic to fetch the proper file from `heroku/languages-github-actions` and tested it to verify it will fetch from the `main` branch or a custom one if the `languages_cli_branch` input is supplied.

.github/workflows/_buildpacks-release.yml
```diff
@@ -75,10 +75,16 @@ jobs:
...
-        # or at least add a CI check to ensure the two versions are in sync.
-        run: cargo install --locked libcnb-cargo@0.13.0
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          LOCKFILE="${{ runner.temp }}/Cargo-$RANDOM.lock"
+          gh api -H "Accept: application/vnd.github.v3.raw" "/repos/heroku/languages-github-actions/contents/Cargo.lock?ref=${{ inputs.languages_cli_branch }}" > $LOCKFILE
+          LIBCNB_PACKAGE_VERSION=$(yq -ptoml -oyaml '.package[] | select(.name == "libcnb-package") | .version' $LOCKFILE)
```
edmorley, line 86: It seems this could be simplified:
- the file could be download from GitHub's raw URL saving needing to use the GitHub CLI along with custom headers/URL (something that people will be much less familiar with). The URL can be found by clicking the "raw" link in GitHub's UI, and for this file is eg: https://raw.githubusercontent.com/heroku/languages-github-actions/main/Cargo.lock
- the file contents can be piped straight to yq (since it supports reading from stdin: https://mikefarah.gitbook.io/yq/#quick-usage-guide) saving having to write to a temporary file

eg something like this:

```suggestion
          LOCKFILE_URL="https://raw.githubusercontent.com/heroku/languages-github-actions/${{ inputs.languages_cli_branch }}/Cargo.lock"
          LIBCNB_PACKAGE_VERSION=$( \
              curl --silent --show-error --fail --retry 5 --retry-all-errors --connect-timeout 10 --max-time 60 "${LOCKFILE_URL}" \
              | yq -ptoml -oyaml '.package[] | select(.name == "libcnb-package") | .version' \
          )
```

edmorley (approved):

Malax (approved):

PR #141, opened 2023-09-25
Use curl best practices for CNB registry check

### LINKED ISSUE #139: Make CNB registry existence check use curl best practices
Curl is invoked here:
https://github.com/heroku/languages-github-actions/blob/f27dde0968647f4c71b0116ab80430621aee9e97/.github/workflows/_buildpacks-release.yml#L251

However, the curl usage is missing a few best practices:
- errors aren't shown
- no retries
- no timeouts
- it manually compares to HTTP 200 instead of using `--fail` and exit codes

It could also likely use a HEAD request instead of a GET, since it doesn't use the response body.

GUS-W-14108732.

colincasey:
Switched `curl` invocation to use head request, handle retries on error, return exit code, etc.

Fixes #139

### DIFF
--- a/.github/workflows/_buildpacks-release.yml
+++ b/.github/workflows/_buildpacks-release.yml
@@ -253,9 +253,12 @@ jobs:
       - name: Check if version is already in the registry
         id: check
         run: |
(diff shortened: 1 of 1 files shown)


edmorley (approved):
Thank you :-)

PR #142, opened 2023-09-25
Adds `Changed` header to buildpack version updates

### LINKED ISSUE #130: Generated changelog for composite buildpacks is missing `### Changed` heading
The CNB release automation creates changelog entries for composite buildpacks explaining the change in versions of the buildpacks within the composite buildpack.

However, currently these changelog entries do not confirm to the https://keepachangelog.com/ style, even though the template changelog header says they do.

Specifically, they are missing the `### Changed` heading:
https://github.com/heroku/buildpacks-nodejs/blob/040e3fbf073cca1838b24e60f5918b666164db72/meta-buildpacks/nodejs/CHANGELOG.md?plain=1#L12-L19

GUS-W-13986595.

colincasey:
- will merge the updated buildpack versions with any existing `Changed` entries
- otherwise will create a new `Changed` header containing the updated buildpack versions

Fixes #130

### DIFF
--- a/src/commands/prepare_release/command.rs
+++ b/src/commands/prepare_release/command.rs
@@ -328,11 +328,11 @@ fn promote_changelog_unreleased_to_version(
     let changes_with_dependencies = (&changelog.unreleased, &updated_dependencies_text);
 
     let body = if let (Some(changes), Some(dependencies)) = changes_with_dependencies {
-        format!("{}\n{}", changes.trim_end(), dependencies)
+        merge_existing_changelog_entries_with_dependency_changes(changes, dependencies)
     } else if let (Some(changes), None) = changes_with_dependencies {
         changes.clone()
     } else if let (None, Some(dependencies)) = changes_with_dependencies {
-        dependencies.clone()
+        format!("### Changed\n\n{dependencies}")
     } else {
         "- No changes.".to_string()
     };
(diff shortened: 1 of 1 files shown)


Malax (approved):
I would've loved to see use working on the AST of the changelog instead of juggling strings. Especially since we already have a parsed representation for the changelog in the `Changelog` struct. We could extend it to include sub-headings. We could be more strict also since we know which sub-headings would be valid in the keep a changelog spec.

I will not block the PR on this (hence the approve). Please either rework the PR to level-up the `Changelog` struct or add an issue for it. :)

colincasey:
@Malax I agree. My first iteration was an attempt to build out a more compliant structure for the Keep A Changelog spec but that's complicated by the fact that several of our existing changelogs do not comply with a stricter spec.  E.g.;
- https://github.com/heroku/buildpacks-jvm/blob/main/buildpacks/jvm/CHANGELOG.md
- https://github.com/heroku/buildpacks-nodejs/blob/main/buildpacks/nodejs-corepack/CHANGELOG.md

To do this levelling-up we would need to first discuss the following:
- How should we deal with those existing non-compliant changelogs?
- How should we track the "No Changes" case which does easily fit into the spec?

I'll create an issue to track that discussion.

colincasey:
Created https://github.com/heroku/languages-github-actions/issues/153

PR #143, opened 2023-09-25
Upgrade `libcnb.rs` to `0.15.0`

### LINKED ISSUE #125: Upgrade to libcnb 0.14.0 and remove hardcoded directory paths

colincasey:
The following general improvements were added for all commands:
- Added helpers for determining the working directory if the optional `--working-directory` argument is given
- Added `Display` implementations for errors shared by several commands (i.e.; `FindReleasableBuildpacksError`, `SetActionOutputError`, `GetWorkingDirectoryError`, `ReadBuildpackDescriptor`)

### `generate_buildpack_matrix`

This command has been updated to remove hardcoded target output directories and instead accept a package directory (`--package-dir`). The other (optional) arguments are to line up with the arguments used by `cargo libcnb package` with sensible defaults.

The `_buildpacks_release.yml` workflow has also been updated to pass through the `--package-dir` argument which is set as the default value `./packaged`.

### `generate_changelog`

Slight modifications to support `libcnb` changes and refactorings detailed above.

### `prepare_release`

Slight modifications to support `libcnb` changes and refactorings detailed above.

### `update_builder`

Slight modifications to support `libcnb` changes and refactorings detailed above.

Fixes #125

### DIFF
--- a/src/buildpacks.rs
+++ b/src/buildpacks.rs
@@ -1,5 +1,7 @@
+use libcnb_common::toml_file::{read_toml_file, TomlFileError};
 use libcnb_data::buildpack::BuildpackDescriptor;
-use libcnb_package::{find_buildpack_dirs, GenericMetadata};
+use libcnb_package::find_buildpack_dirs;
+use std::fmt::{Display, Formatter};
 use std::path::{Path, PathBuf};
 use std::process::{Command, ExitStatus};
 
@@ -26,11 +28,11 @@ pub(crate) fn calculate_digest(digest_url: &String) -> Result<String, CalculateD
 }
 
 pub(crate) fn read_image_repository_metadata(
-    buildpack_descriptor: &BuildpackDescriptor<GenericMetadata>,
+    buildpack_descriptor: &BuildpackDescriptor,
 ) -> Option<String> {
     let metadata = match buildpack_descriptor {
-        BuildpackDescriptor::Single(descriptor) => &descriptor.metadata,
-        BuildpackDescriptor::Meta(descriptor) => &descriptor.metadata,
+        BuildpackDescriptor::Component(descriptor) => &descriptor.metadata,
+        BuildpackDescriptor::Composite(descriptor) => &descriptor.metadata,
     };
 
     #[allow(clippy::redundant_closure_for_method_calls)]
@@ -42,20 +44,78 @@ pub(crate) fn read_image_repository_metadata(
         .map(|value| value.to_string())
 }
 
-pub(crate) fn find_releasable_buildpacks(starting_dir: &Path) -> std::io::Result<Vec<PathBuf>> {
-    find_buildpack_dirs(starting_dir, &[starting_dir.join("target")]).map(|results| {
-        results
-            .into_iter()
-            .filter(|dir| dir.join("CHANGELOG.md").exists())
-            .collect()
-    })
+pub(crate) fn find_releasable_buildpacks(
+    starting_dir: &Path,
+) -> Result<Vec<PathBuf>, FindReleasableBuildpacksError> {
+    find_buildpack_dirs(starting_dir)
+        .map(|results| {
+            results
+                .into_iter()
+                .filter(|dir| dir.join("CHANGELOG.md").exists())
+                .collect()
+        })
+        .map_err(|e| FindReleasableBuildpacksError(starting_dir.to_path_buf(), e))
+}
+#[derive(Debug)]
+pub(crate) struct FindReleasableBuildpacksError(PathBuf, ignore::Error);
+
+impl Display for FindReleasableBuildpacksError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let path = &self.0;
+        let error = &self.1;
+        write!(
+            f,
+            "I/O error while finding buildpacks\nPath: {}\nError: {error}",
+            path.display()
+        )
+    }
+}
+
+pub(crate) fn read_buildpack_descriptor(
+    dir: &Path,
+) -> Result<BuildpackDescriptor, ReadBuildpackDescriptorError> {
+    let buildpack_path = dir.join("buildpack.toml");
+    read_toml_file::<BuildpackDescriptor>(&buildpack_path)
+        .map_err(|e| ReadBuildpackDescriptorError(buildpack_path, e))
+}
+
+#[derive(Debug)]
+pub(crate) struct ReadBuildpackDescriptorError(PathBuf, TomlFileError);
+
+impl Display for ReadBuildpackDescriptorError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let buildpack_path = &self.0;
+        let error = &self.1;
+        match error {
+            TomlFileError::IoError(source) => {
+                write!(
+                    f,
+                    "Failed to read buildpack\nPath: {}\nError: {source}",
+                    buildpack_path.display()
+                )
+            }
+            TomlFileError::TomlDeserializationError(source) => {
+                write!(
+                    f,
+                    "Failed to deserialize buildpack\nPath: {}\nError: {source}",
+                    buildpack_path.display()
+                )
+            }
+            TomlFileError::TomlSerializationError(source) => {
+                write!(
+                    f,
+                    "Failed to serialize buildpack\nPath: {}\nError: {source}",
+                    buildpack_path.display()
+                )
+            }
+        }
+    }
 }
 
 #[cfg(test)]
 mod test {
     use crate::buildpacks::read_image_repository_metadata;
     use libcnb_data::buildpack::BuildpackDescriptor;
-    use libcnb_package::GenericMetadata;
 
     #[test]
     fn test_read_image_repository_metadata() {
@@ -73,8 +133,7 @@ id = "*"
 repository = "repository value"
 "#;
 
-        let buildpack_descriptor =
-            toml::from_str::<BuildpackDescriptor<GenericMetadata>>(data).unwrap();
+        let buildpack_descriptor = toml::from_str::<BuildpackDescriptor>(data).unwrap();
         assert_eq!(
             read_image_repository_metadata(&buildpack_descriptor),
             Some("repository value".to_string())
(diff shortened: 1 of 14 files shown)


src/buildpacks.rs
```diff
@@ -42,20 +44,78 @@ pub(crate) fn read_image_repository_metadata(
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let path = &self.0;
+        let error = &self.1;
+        write!(
+            f,
+            "I/O error while finding buildpacks\nPath: {}\nError: {error}",
+            path.display()
+        )
+    }
+}
```
Malax, line 72: Couldn't we use `thiserror` here? We wouldn't need to implement `Display` by hand (which requires a bunch of boilerplate as seen here) and we would also get [std::error::Error](https://doc.rust-lang.org/std/error/trait.Error.html) implementations "for free". This applies to all places where we added `Display` implementations.

src/commands/mod.rs
```diff
@@ -1,4 +1,34 @@
...
+            if let Some(path) = value {
+                if path.is_absolute() {
+                    path
+                } else {
+                    current_dir.join(path)
+                }
+            } else {
+                current_dir
+            }
+        })
```
Malax, line 23: Just as a side-note: I think this implementation would be more readable in this form:

```rust
match value {
    None => std::env::current_dir(),
    Some(path) if path.is_absolute() => Ok(path),
    Some(path) => std::env::current_dir().map(|current_dir| current_dir.join(path))
}.map_err(ResolvePathError)
```

Ideal functions (pure functions) are just mappings from the input values to the output values. In simple cases like this where we can actually spell out the mappings explicitly, `match` is a good tool for doing so, even when the function is not pure like this one. I'm mentioning the mapping analogy as it has been a useful mental shortcut for me when thinking about code. :)

src/commands/mod.rs
```diff
@@ -1,4 +1,34 @@
+use std::fmt::{Display, Formatter};
+use std::io;
+use std::path::PathBuf;
+
 pub(crate) mod generate_buildpack_matrix;
 pub(crate) mod generate_changelog;
 pub(crate) mod prepare_release;
 pub(crate) mod update_builder;
+
+pub(crate) fn resolve_path(value: Option<PathBuf>) -> Result<PathBuf, ResolvePathError> {
```
Malax, line 10: When a function only takes an optional argument I immediately suspect something is off with the design since there are only very few use-cases where that signature makes logical sense. In many cases it's hiding an implicit default deep in the code.

This is the case here too. Since the `working_dir` field of the user-provided args is optional it must've been more convenient to design `resolve_path` so that it can just take that value verbatim. However, this moves the _business logic_ of "when no working dir is specified, the current dir is used" down into a utility function - entangling these concerns across the call stack.

In addition, there are call-sites where the `value` passed in is not an `Option`. This results in the need to add `Some()` at the call site which serves no logical purpose in that context.

We should move the implicit default up where it belongs - into `GenerateBuildpackMatrixArgs`. Adding a function to `GenerateBuildpackMatrixArgs` that does the resolution of the path would go a long way. I imagine the signature to look something like this: `fn working_dir(&self) -> std::io::Result<PathBuf>`.

With the default value concern out of the way, we can refactor `resolve_path`. First, we can rename it to something like `absolute_path` and change the signature to something like `fn absolute_path(path: &Path, base: &Path) -> PathBuf`. This would make the function pure (with the immediate benefit that is much easier to test).

Where did the error go? Well, the only case where this can fail is if we need to determine current directory. We moved this out of this function to where we actually need to know the current directory - when resolving the working directory specified by the user. This means that at call-sites where we always passe `Some(foo)` the need to handle an error that could actually never logically occur goes away entirely!

I think this is an excellent case study of the impact of designing functions so that they make sense in a vacuum. This has positive ripple effects throughout the code that go way beyond the function itself. In the case here:

- Readers looking at `GenerateBuildpackMatrixArgs` can now immediately see what the default value for `working_dir` is.
- `resolve_path`/`absolute_path` is now pure and easily testable because there are no side-effects.
- Users of `resolve_path`/`absolute_path` don't need to deal with errors when no error can actually occur.
- We can remove the `ResolvePathError` struct.

Malax (changes requested):

src/buildpacks.rs
colincasey: switched all the `Display` implementations for `thiserror`

src/commands/mod.rs
colincasey: This one came out of the desire to have several of the command args (`GenerateBuildpackMatrixArgs`, `GenerateChangelogArgs`, and `PrepareReleaseArgs`) allow for the option to change their working directory or default to the current one. With no easy way to handle this as a default in `clap` the function started out with the `Option<PathBuf>` for convenience. Then later, when I needed almost the exact same functionality with a non-optional path I just repurposed the function instead of refining it.

I've broken down the single `resolve_path` function which is currently doing too much into the following:
- a pure `resolve_path(path: &Path, base: &Path) -> PathBuf` function that, as you say, makes sense in a vacuum and does not error
- a convenience `resolve_working_dir_from_current_dir(value: Option<PathBuf>) -> std::io::Result<PathBuf>` function that handles resolving the optional `working_dir` argument for the various `clap` commands

I've also dropped the `ResolvePathError`.

Malax: I'm not sure I understand. If we'd add functions to get either the explicitly configured working directory value or the default (`std::env::current_dir`) to `GenerateBuildpackMatrixArgs`, `GenerateChangelogArgs`, and `PrepareReleaseArgs`, we can work with `resolve_path` in all cases and don't need a highly specific helper (plus the need to come up with a name for it) that does hide the default far away from where it should be logically placed.

colincasey: I think you're misunderstanding the intent. But, as it stands, I don't want to hold up this PR for a minor detail so I've removed the function.

src/commands/generate_buildpack_matrix/command.rs
```diff
@@ -1,34 +1,70 @@
...
+    pub(crate) release: Option<bool>,
+    #[arg(long, default_value = "x86_64-unknown-linux-musl")]
+    pub(crate) target: String,
+}
+
+pub(crate) fn execute(args: GenerateBuildpackMatrixArgs) -> Result<()> {
+    let working_dir = std::env::current_dir()
+        .map(|base| match args.working_dir {
+            Some(path) => resolve_path(&path, &base),
+            None => base,
```
edmorley, line 35: The match here could be replaced by `args.working_dir.map_or(...)`

(and similarly for the other usages of this pattern in the other `execute()` methods)

Malax (approved):

src/commands/generate_buildpack_matrix/command.rs
```diff
@@ -1,34 +1,70 @@
...
+    let working_dir = std::env::current_dir()
+        .map(|base| {
+            args.working_dir
+                .map_or(base.clone(), |path| resolve_path(&path, &base))
+        })
+        .map_err(Error::ResolveWorkingDir)?;
+
+    let package_dir = resolve_path(&args.package_dir, &working_dir);
+
+    let cargo_profile = if args.release.unwrap_or(true) {
```
edmorley, line 41: I think this is a bug? It should be `unwrap_or(false)` rather than `unwrap_or(true)`, since at the moment passing `--release` to the CLI gives a release build, but not passing it also gives a release build.

Either we should drop the `release` arg, and only support release builds here, or else default to non-release, and have `--release` as the way to specify release. 

I think my preference is just to drop the `release` are entirely -- this release automation is all about producing a final release product, not a development build.

Also, if we did want to support a `--release` arg, then we perhaps would instead want to change `args.release` from `Option<bool>` to `bool`, and instead set a Clap default value on the `GenerateBuildpackMatrixArgs` struct directly.

src/commands/generate_buildpack_matrix/command.rs
```diff
@@ -1,34 +1,70 @@
...
+use std::path::{Path, PathBuf};
 
 type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Parser, Debug)]
 #[command(author, version, about = "Generates a JSON list of buildpack information for each buildpack detected", long_about = None)]
-pub(crate) struct GenerateBuildpackMatrixArgs;
+pub(crate) struct GenerateBuildpackMatrixArgs {
+    #[arg(long)]
+    pub(crate) working_dir: Option<PathBuf>,
```
edmorley, line 22: This PR adds a new `--working-dir` argument to most of the CLI commands (for some it's a rename of `--path`), however nothing seems to set it (e.g. from the GitHub Actions workflow), so it seems unused?

Whilst for end-user tooling (such as `cargo libcnb`) having an arg to override the working directory might be useful, for the CNB release automation in this repo, we're the only consumers, so if we're not using an arg then we might as well drop it. (We can always add it back later if needed.)

Dropping these unused args would also simplify a lot of the debate over the path resolving functionality discussed in:
https://github.com/heroku/languages-github-actions/pull/143#discussion_r1347103266

edmorley (changes requested):

src/commands/generate_buildpack_matrix/command.rs
edmorley: Addressed in 4ff35503f2f9ef293637e482c7d047c3b4e77544

src/commands/generate_buildpack_matrix/command.rs
edmorley: After speaking with Manuel, he agreed, so I've gone ahead and addressed this in c7c3dd4ce258a21bc0641dcabd87dd21bdb978b0 to speed things up.

Between that commit and the `--release` removal one, the PR has now reduced in size from `+527 −605` to `+490 −601`.

edmorley:
@colincasey Thank you for fixing the lint error - it was on my list but I got pulled away for the adhoc stack image update.

edmorley (approved):

PR #183, opened 2024-01-24
Prepare release workflow for classic buildpacks

colincasey:
Add a GitHub Actions workflow for preparing a new "classic" buildpack release. This is based on the workflow that @edmorley added for python (https://github.com/heroku/heroku-buildpack-python/pull/1506).

> [!NOTE]
> This workflow may not be compatible with all of our classic buildpacks such as Ruby or PHP which do not follow the conventions being automated here.

### DIFF
--- a/.github/workflows/_classic-buildpack-prepare-release.yml
+++ b/.github/workflows/_classic-buildpack-prepare-release.yml
@@ -1,0 +1,117 @@
+name: Prepare release
+
+on:
+  workflow_call:
+    inputs:
+      app_id:
+        description: Application ID of GitHub application (e.g.; the Linguist App)
+        type: string
+        required: true
+      app_email:
+        description: The email address of the GitHub application bot user (e.g.; the Linguist App)
+        type: string
+        required: false
+        default: ${{ vars.LINGUIST_GH_APP_EMAIL }}
+      app_username:
+        description: The username of the GitHub application bot user (e.g.; the Linguist App)
+        type: string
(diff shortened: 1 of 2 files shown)


joshwlewis (approved):
Looks awesome!

.github/workflows/_classic-buildpack-prepare-release.yml
```diff
@@ -0,0 +1,117 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          # We always want the version bump/changelog and resultant PR to target main, not the branch of the workflow_dispatch.
+          ref: main
+          # Using the GH application token here will configure the local git config for this repo with credentials
+          # that can be used to make signed commits that are attributed to the GH application user
+          token: ${{ steps.generate-token.outputs.app_token }}
+
+      - name: Determine existing published version
```
edmorley, line 67: I have an alternative approach in #184, which uses git tags to calculate the existing version, and so works with Ruby and PHP too.

I've also added support for running a custom command before committing, so the Ruby buildpack can make it's additional edits. (Such as updating the buildpack version in [`language_pack/version.rb`](https://github.com/heroku/heroku-buildpack-ruby/blob/main/lib/language_pack/version.rb) plus moving the `changelogs/unreleased/` directory to `changelogs/vNNN/` etc)

colincasey:
closing this in favor of #184 which includes several improvements

PR #194, opened 2024-02-05
Multi-arch buildpack automation

joshwlewis:


.github/workflows/_buildpacks-release.yml
```diff
@@ -177,26 +201,103 @@ jobs:
...
-      - name: Tag and publish buildpack
-        if: inputs.dry_run == false && steps.check.outputs.published_to_docker == 'false'
+      - name: Publish to Docker Hub
+        if: steps.check.outputs.published_to_docker == 'false'
+        env:
+          TARGETS: ${{ toJSON(matrix.targets) }}
         run: |
-          docker tag ${{ matrix.buildpack_id }} ${{ matrix.docker_repository }}:${{ matrix.buildpack_version }}
-          docker push ${{ matrix.docker_repository }}:${{ matrix.buildpack_version }}
+          temp_tag="${{ matrix.docker_repository }}:_${{ github.run_id }}"
```
colincasey, line 222: One thing I tried to use the Rust-based commands in this repo for was to offload any sufficiently complex script into something that could be easier to maintain and test.  This seems like good candidates for a new custom command and, if you were to do that, the subsequent temp tag cleanup could probably be rolled into that as well.

.github/workflows/_buildpacks-release.yml
```diff
@@ -71,14 +71,33 @@ jobs:
...
+        env:
+          RUST_TRIPLES: ${{ steps.generate-buildpack-matrix.outputs.rust_triples }}
+        run: |
+          for triple in $(jq --exit-status -r '.[]' <<< "${RUST_TRIPLES}"); do
+            if [[ "$triple" == "aarch64-unknown-linux-musl" ]]; then
+              sudo apt-get install musl-tools gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libc6-dev-arm64-cross --no-install-recommends
+            elif [[ "$triple" == "x86_64-unknown-linux-musl" ]]; then
+              sudo apt-get install musl-tools --no-install-recommends
+            else
+              echo "WARNING: Unknown triple: $triple. Cross-compile dependencies were skipped."
```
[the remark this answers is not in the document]
joshwlewis: Mostly for any bash buildpacks (`heroku/nodejs-npm`). They'll have a target of `linux` (with no architecture), so we won't have a rust triple for that. We don't need a rust triple for that either, since we don't run `cargo libcnb package` on it.

edmorley (commented):

Longer term (some time after we have Heroku-24 multi-arch working overall, and upstream Pack supports packaging multi-arch natively) I'd love for us to push down most of the Rust implementation parts in this repo into libcnb package (or another subcommand) itself. 

But at a high level this looks great to get us started :-)
(Happy to defer the rest of the review to Colin)

.github/workflows/_buildpacks-release.yml
joshwlewis: Yeah, that's a good point. Though, I'm not super eager to rewrite it because:

1) I don't think the lifetime of this code is very long. When CNB support for multiarch launches and libcnb.rs supports that, a good chunk of this will be handled by `pack` and `cargo libcnb package`. 
2) There's a half a dozen or more command line calls to model in rust.

Though I think I'll take a stab at moving a few more calculations (like say `perm_tag` / `temp_tag`) into the rust matrix.

.github/workflows/_buildpacks-release.yml
```diff
@@ -201,6 +201,9 @@ jobs:
       - name: Install Pack CLI
         uses: buildpacks/github-actions/setup-pack@v5.5.4
 
+      - name: Install Crane
+        uses: uses: imjasonh/setup-crane@v0.3
```
edmorley, line 205: We should be cautious about using Actions from third parties, particularly when they don't seem to be very popular repos, and the Action has access to our most sensitive credentials (such as the Docker Hub tokens).

For installing crane elsewhere in our automation, we use the `setup-tools` Action from upstream CNB:
https://github.com/heroku/languages-github-actions/blob/d3dcac71c07fcb38eee51d53a9b861d7abffe0e3/.github/workflows/_buildpacks-release.yml#L258-L259

You'll need to pass a specific `crane-version` if you want newer than the default, since the default version is a bit stale, xref:
https://github.com/buildpacks/github-actions/pull/262

.github/workflows/_buildpacks-release.yml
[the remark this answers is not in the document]
joshwlewis: I've changed the rust logic that generates the list of rust triples. It will now error in any case where a rust triple is expected (libcnb.rs bulidpacks) and the target doesn't have a known triple. This else has been removed.

src/commands/generate_buildpack_matrix/command.rs
```diff
@@ -71,42 +62,123 @@ pub(crate) fn execute(args: &GenerateBuildpackMatrixArgs) -> Result<()> {
...
     Ok(())
 }
 
 pub(crate) fn extract_buildpack_info(
     buildpack_descriptor: &BuildpackDescriptor,
     buildpack_dir: &Path,
-    packaged_buildpack_dir_resolver: &impl Fn(&BuildpackId) -> PathBuf,
-) -> Result<BTreeMap<String, String>> {
+    package_dir: &Path,
+) -> Result<BTreeMap<String, serde_json::Value>> {
```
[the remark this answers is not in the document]
joshwlewis: I've done some refactoring so that these `serde_json::Value` types are no longer used. Instead, I'm using structs that have `Deserialize` and I think it's looking better.

.github/workflows/_buildpacks-release.yml
joshwlewis: Yeah, I vetted this action. While it's not super popular, `imjasonh` is one of the primary devs on `crane` and regularly creates releases. By using `crane` at all, we're already exposed to this same user. The action is also version locked and the action source does exactly what you'd expect. 

I don't have any issue using the buildpacks one either, though. I'll switch it.

.github/workflows/_buildpacks-release.yml
```diff
@@ -60,7 +60,7 @@ env:
 jobs:
   compile:
     name: Compile Buildpacks
-    runs-on: ubuntu-latest
+    runs-on: ${{ inputs.ip_allowlisted_runner }}
```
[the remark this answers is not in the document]
joshwlewis: This came along from an old branch @colincasey started and maybe the reason it was selected is no longer present. It looks like `ubuntu-latest` works fine, so going with that.

.github/workflows/_buildpacks-release.yml
```diff
@@ -94,62 +111,31 @@ jobs:
...
-
-      - name: Generate buildpack matrix
-        id: generate-buildpack-matrix
-        run: actions generate-buildpack-matrix --package-dir ${{ env.PACKAGE_DIR }}
-
-      - name: Generate changelog
-        id: generate-changelog
-        run: actions generate-changelog --version ${{ steps.generate-buildpack-matrix.outputs.version }}
-
-      - name: Temporary fix for bash-based buildpacks
```
colincasey, line 115: ngl, it's nice to see this gone 😅

.github/workflows/_buildpacks-release.yml
```diff
@@ -177,26 +163,90 @@ jobs:
...
+          target_temp_tags=($(jq --exit-status -r "map(.temporary_tag) | join(\" \")" <<< "${TARGETS}"))
+          for i in "${!target_temp_tags[@]}"; do
+            perm_tag=$(jq --exit-status -r ".[$i].permanent_tag" <<< "${TARGETS}")
+            crane copy "${target_temp_tags[i]}" "${perm_tag}"
+          done
+          # promote primary image manifest or manifest list to permanent tag
+          crane copy "${{ matrix.temporary_tag }}" "${{ matrix.permanent_tag }}"
+
+      - name: Unpublish temp tags from this run
+        if: always()
```
colincasey, line 222: This shouldn't run if `steps.check.outputs.published_to_docker == 'false'` as can be seen here: https://github.com/heroku/buildpacks-nodejs/actions/runs/8712758197/job/23899957917.

Or, if it always runs, it shouldn't error out like it does in the run linked above.

.github/workflows/_buildpacks-release.yml
```diff
@@ -94,62 +111,31 @@ jobs:
...
-                  if [ "${bash_buildpack_source_dir}" = "${dep_path}" ]; then
-                    echo "replacing ${dep} with ${bash_buildpack_output_dir}"
-                    sed -i 's|'"$dep"'|'"$bash_buildpack_output_dir"'|g' "${output_dir}/package.toml"
-                  fi
-                done
+            cd "$buildpack_dir"
+            for target in $(jq --exit-status -c '.targets | .[]' <<< "${buildpack}"); do
+              output_dir=$(jq --exit-status -r '.output_dir' <<< "${target}")
+              if [[ "$buildpack_dir" -ef "$output_dir" ]]; then
+                echo "Skipping packaging of ${output_dir}. Not a libcnb.rs buildpack."
```
colincasey, line 126: hmm, @joshwlewis just wondering if this is the correct action to take here. We still need to move the bash buildpacks into the `${package_dir}` location because at the end of this Compile Buildpacks job we cache the contents of the `${package_dir}` folder.  Only the `${package_dir}` contents are restored during the publishing steps (we don't do a git checkout) so, however this works, the bash buildpack should be copied into that folder.

src/commands/generate_buildpack_matrix/command.rs
```diff
@@ -71,42 +82,267 @@ pub(crate) fn execute(args: &GenerateBuildpackMatrixArgs) -> Result<()> {
...
+        .flat_map(read_buildpack_targets)
+        .filter_map(|t| rust_triple(&t).ok())
+        .collect::<HashSet<String>>();
+
+    actions::set_output(
+        "rust_triples",
+        serde_json::to_string(&rust_triples).map_err(Error::SerializingJson)?,
+    )
+    .map_err(Error::SetActionOutput)?;
+
```
colincasey, line 96: @joshwlewis We should provide [summaries](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary) for these outputs.  E.g.; pretty-print the json into a markdown code block or format it into a table. This would be useful for communicating what's being processed (especially now that multiple targets are involved).

It would be easy to follow a similar pattern as `actions::set_output` with a `actions::set_summary` function that writes to `$GITHUB_STEP_SUMMARY`.

.github/workflows/_buildpacks-release.yml
```diff
@@ -177,26 +163,90 @@ jobs:
...
+          TARGETS: ${{ toJSON(matrix.targets) }}
+        run: |
+          # Promote target temp tags to stable tags
+          target_temp_tags=($(jq --exit-status -r "map(.temporary_tag) | join(\" \")" <<< "${TARGETS}"))
+          for i in "${!target_temp_tags[@]}"; do
+            perm_tag=$(jq --exit-status -r ".[$i].permanent_tag" <<< "${TARGETS}")
+            crane copy "${target_temp_tags[i]}" "${perm_tag}"
+          done
+          # promote primary image manifest or manifest list to permanent tag
+          crane copy "${{ matrix.temporary_tag }}" "${{ matrix.permanent_tag }}"
```
colincasey, line 219: Could we provide a [job summary](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary) here that links to the published docker image?

.github/workflows/_buildpacks-release.yml
```diff
@@ -218,9 +268,11 @@ jobs:
...
         run: |
           for buildpack in $(jq --exit-status -c '.[]' <<< '${{ needs.compile.outputs.buildpacks }}'); do
-            artifact_prefix=$(jq --exit-status -r '.buildpack_artifact_prefix' <<< "${buildpack}")
-            output_dir=$(jq --exit-status -r '.buildpack_output_dir' <<< "${buildpack}")
-            pack buildpack package "${artifact_prefix}.cnb" --config "${output_dir}/package.toml" --format file --verbose
+            for target in $(jq --exit-status -c ".targets | .[]" <<< "${buildpack}"); do
+              output_dir=$(jq --exit-status -r ".output_dir" <<< "${target}")
+              cnb_file=$(jq --exit-status -r ".cnb_file" <<< "${target}")
+              pack buildpack package "$cnb_file" --config "${output_dir}/package.toml" --format file --verbose
+            done
```
colincasey, line 275: Could we provide a [job summary](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary) somewhere around here that links to the github release?

.github/workflows/_buildpacks-release.yml
```diff
@@ -279,7 +331,7 @@ jobs:
           token: ${{ secrets.cnb_registry_token }}
           id: ${{ matrix.buildpack_id }}
           version: ${{ matrix.buildpack_version }}
-          address: ${{ matrix.docker_repository }}@${{ steps.digest.outputs.value }}
+          address: ${{ matrix.image_repository }}@${{ steps.digest.outputs.value }}
```
colincasey, line 334: Could we provide a [job summary](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary) here that links to the new CNB registry entry?

.github/workflows/_buildpacks-release.yml
```diff
@@ -279,7 +331,7 @@ jobs:
...
 
   update-builder:
```
colincasey, line 336: Could we provide a [job summary](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary) here that links to the builder PR that gets created?

.github/workflows/_buildpacks-release.yml
joshwlewis: > This shouldn't run if `steps.check.outputs.published_to_docker == 'false'`

I think it should run. In both dry run mode and scenarios where we've already published the versioned tag (`heroku/buildpack-procfile:3.0.0`), we still want to make sure the temp tags are deleted (`heroku/buildpack-procfile:_12345`). We never want to leave these around. If any of the tags don't exist (say due to a previous error), the deletion step still exits with a success.

That run errored out due to insufficient scope - which means the docker personal access token doesn't have `delete`. Since we're deleting tags, each repo using this action will need new personal access tokens with `delete` scope. I think that scenario should be an error - it's leaving extra tags around and needs developer intervention.

.github/workflows/_buildpacks-release.yml
joshwlewis: Hmm, I see. Well, I could restore that code. Or, we could run `git checkout` on the other steps. Neither is as efficient as I'd like.

Another option maybe - restore the copy logic, but instead of rewriting other package.tomls, I could symlink the bash buildpack source directory to it's location in the package dir.

.github/workflows/_buildpacks-release.yml
colincasey: Ah, my bad. I thought it was erroring out because there were no tags to delete. Ignore my earlier comment 😅

.github/workflows/_buildpacks-release.yml
colincasey: Yeah, this annoying script solved 2 problems:
1. persisting the bash buildpack so you have something to package in later jobs
2. rewriting package.toml files that reference the bash buildpack so that composite buildpacks can be packaged

If you can accommodate those two cases via some other mechanism that would be good. Otherwise we'll still need this kludge around for awhile longer.

.github/workflows/_buildpacks-release.yml
joshwlewis: Good idea. But I think I'll decline that for this PR. I'm not making significant changes to this step and this PR is already massive.

.github/workflows/_buildpacks-release.yml
joshwlewis: Good idea. But I think I'll decline that for this PR. I'm not making significant changes to this step and this PR is already massive. I did add job summaries for matrix generation, temporary tags, and stable tags.

.github/workflows/_buildpacks-release.yml
joshwlewis: I left the copy in place. I dropped the dependency rewrite stuff, though. Instead, I used the `Checkout` action, but only for composite buildpacks. That way, all buildpacks are in the package dir, and any composite buildpacks can find bash buildpacks in their original spot.

src/commands/generate_buildpack_matrix/command.rs
joshwlewis: Done!

.github/workflows/_buildpacks-release.yml
joshwlewis: Done!

.github/workflows/_buildpacks-release.yml
```diff
@@ -177,32 +174,108 @@ jobs:
...
+            fi
+          done
+          exit $status
 
   publish-github:
     name: Publish → GitHub Release
     needs: [compile]
     runs-on: ${{ inputs.ip_allowlisted_runner }}
     steps:
+      - name: Checkout
```
edmorley, line 274: Should this Checkout step have a `matrix.buildpack_type == 'composite'` conditional, like the Checkout step under the `publish-docker` job?

.github/workflows/_buildpacks-release.yml
edmorley, line 1: This is not directly related to this change, but I was wondering what you thought about renaming the `publish-cnb` job to eg `publish-registry` (or `publish-cnb-registry`)?

Every time I read the `publish-cnb` name I think "generate and publish the .cnb files" which is not what that step does.

src/commands/generate_buildpack_matrix/errors.rs
```diff
@@ -18,7 +19,15 @@ pub(crate) enum Error {
...
     FixedVersion(HashSet<String>),
     #[error(transparent)]
-    SetActionOutput(SetActionOutputError),
+    WriteActionData(WriteActionDataError),
+    #[error("Unknown target configuration. Couldn't determine a rust triple for {0:?}.")]
+    UnknownRustTarget(BuildpackTarget),
+    #[error("Couldn't determine buildpack type. Found evidence for two or more buildpack types (bash, composite, libcnb.rs) in {0}.")]
+    MultipleTypes(PathBuf),
+    #[error(
+        "Couldn't determine buildpack type. Found no evidence of a bash, composite, or libccnb.rs buildpack in {0}."
```
edmorley, line 28:
```suggestion
        "Couldn't determine buildpack type. Found no evidence of a bash, composite, or libcnb.rs buildpack in {0}."
```

edmorley (commented):
Thank you for getting this working!

.github/workflows/_buildpacks-release.yml
joshwlewis: Yeah, I've had similar thoughts. I'll rename.

.github/workflows/_buildpacks-release.yml
edmorley: Ah of course, it handles multiple buildpacks in the one step. Agree not worth it

edmorley: (Also hopefully we can just remove Bash buildpack support in the near future, if we can wean functions off of the old bash buildpack... the warning about needing a lockfile has been showing for quite some time now?)

src/commands/generate_buildpack_matrix/command.rs
```diff
@@ -19,41 +20,55 @@ type Result<T> = std::result::Result<T, Error>;
...
         .collect::<Result<Vec<_>>>()?;
 
-    let includes_json = serde_json::to_string(&includes).map_err(Error::SerializingJson)?;
+    let buildpacks_json =
+        serde_json::to_string_pretty(&buildpacks_info).map_err(Error::SerializingJson)?;
 
-    actions::set_output("buildpacks", includes_json).map_err(Error::SetActionOutput)?;
+    actions::set_output("buildpacks", &buildpacks_json).map_err(Error::WriteActionData)?;
+    actions::set_summary(format!(
+        "<details><summary>Buildpack Matrix</summary>\n\n```json\n{buildpacks_json}\n```\n</details>"
```
colincasey, line 69: nice to get visibility into this data finally! and good call putting this into a summary table. the output for https://github.com/heroku/buildpacks-nodejs/actions/runs/8926879828 is huge 😅 

in the future, a flattened HTML table might be a reasonable UX improvement but let's not worry about that now.

colincasey (approved):
Looks good!

src/commands/generate_buildpack_matrix/command.rs
joshwlewis: Yeah, it's huge. Obviously, the Node.js CNB will be huge since it includes so many buildpacks, and 2x targets for each. Other buildpacks should be a little easier to parse.

PR #197, opened 2024-02-12
Drop some unneeded dependencies

edmorley:
* Drops some unused `chrono` features, see: https://github.com/chronotope/chrono/blob/main/README.md#crate-features
* Switches from [rand](https://crates.io/crates/rand) to the lightweight [fastrand](https://crates.io/crates/rand) crate, since we don't need the much larger cryptographically secure library. (Other non-cryptographic crates we use, like `tempfile` use `fastrand` too.)

Reduces the number of transitive dependencies by 20.

GUS-W-15019814.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,21 +12,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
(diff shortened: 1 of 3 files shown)


Malax (approved):
Nice!

PR #212, opened 2024-04-16
Don't move and rewrite bash buildpacks

joshwlewis:
This PR is stacked on top of #194. In that PR, there is a defined property for `.output_dir` on each target. This PR detects bash buildpacks, and uses the buildpack directory for the `.output_dir` in that case. This way, we don't have to move the bash buildpacks around, then rewrite dependencies to that buildpack.


joshwlewis:
I'm going to fold this into #194. Got a bit too wonky to rebase after some other changes.

PR #239, opened 2024-07-04
Restore safe Docker publishing

### LINKED ISSUE #238: The publishing buildpacks to docker step doesn't check if the tag already exists
One of the original design principals of the CNB release automation was that releases be idempotent - ie: it should be possible to safely retrigger a release job and not have it clobber any existing release (but will repeat any steps that haven't yet been performed).

As such, in the original "publish to Docker Hub" step, there was a check to see if the image was already published:
https://github.com/heroku/languages-github-actions/blob/08617454b23d0545ae314bbc7c4e009b06bc344a/.github/workflows/_buildpacks-release.yml#L191-L193

However, after multi-arch support was added, there is now no such check:
https://github.com/heroku/languages-github-actions/blob/e6071f8c22c13174b590fbd32f51a80aee957d37/.github/workflows/_buildpacks-release.yml#L148-L269

As such, the images are `crane copy`d again, even if they exist.

For example see:
https://github.com/heroku/buildpacks-procfile/actions/runs/9782898585/job/27010314681#step:8:42

Now thankfully in that run, the image had the same SHA256, so it didn't end up changing the remote tag. However, I believe that was only by luck due to the Rust compile being somewhat deterministic - so the compiled binaries were bit for bit identical - but I don't believe Rust guarantees that, so we shouldn't rely upon it.

Instead we should either:
1. Use the `--no-clobber` option for the `crane copy` used to promote the tags (see: https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_copy.md) - though it's not clear whether this will result in `crane` exiting non-zero or not (and we want it to exit zero, like other steps in the workflow that skip)
2. Or, have an explicit "does this tag exist" check before running the `crane copy` to promote the tags

cc @colincasey @joshwlewis

colincasey:
This fix restores the safety of triggering the release automation on an existing release which was lost in the updates for publishing multi-arch Docker images in #194.

The buildpack matrix output can contain two variants of images to publish:

- A CNB with a single target. E.g.;

    ```json
    {
      "buildpack_id": "heroku/nodejs-function",
      "buildpack_version": "3.2.6",
      "buildpack_type": "composite",
      "buildpack_dir": "/home/runner/work/buildpacks-nodejs/buildpacks-nodejs/meta-buildpacks/nodejs-function",
      "targets": [
        {
          "os": "linux",
          "arch": "amd64",
          "rust_triple": "x86_64-unknown-linux-musl",
          "oci_target": "linux/amd64",
          "cnb_file": "heroku_nodejs-function.cnb",
          "stable_tag": "docker.io/heroku/buildpack-nodejs-function:3.2.6",
          "temporary_tag": "docker.io/heroku/buildpack-nodejs-function:_9781380433",
          "output_dir": "/home/runner/work/buildpacks-nodejs/buildpacks-nodejs/./packaged/x86_64-unknown-linux-musl/release/heroku_nodejs-function"
        }
      ],
      "image_repository": "docker.io/heroku/buildpack-nodejs-function",
      "stable_tag": "docker.io/heroku/buildpack-nodejs-function:3.2.6",
      "temporary_tag": "docker.io/heroku/buildpack-nodejs-function:_9781380433"
    }
    ```

- Or a CNB with multiple targets. E.g.;

    ```json
    {
      "buildpack_id": "heroku/nodejs-npm-install",
      "buildpack_version": "3.2.6",
      "buildpack_type": "libcnb",
      "buildpack_dir": "/home/runner/work/buildpacks-nodejs/buildpacks-nodejs/buildpacks/nodejs-npm-install",
      "targets": [
        {
          "os": "linux",
          "arch": "amd64",
          "rust_triple": "x86_64-unknown-linux-musl",
          "oci_target": "linux/amd64",
          "cnb_file": "heroku_nodejs-npm-install_linux-amd64.cnb",
          "stable_tag": "docker.io/heroku/buildpack-nodejs-npm-install:3.2.6_linux-amd64",
          "temporary_tag": "docker.io/heroku/buildpack-nodejs-npm-install:_9781380433_linux-amd64",
          "output_dir": "/home/runner/work/buildpacks-nodejs/buildpacks-nodejs/./packaged/x86_64-unknown-linux-musl/release/heroku_nodejs-npm-install"
        },
        {
          "os": "linux",
          "arch": "arm64",
          "rust_triple": "aarch64-unknown-linux-musl",
          "oci_target": "linux/arm64",
          "cnb_file": "heroku_nodejs-npm-install_linux-arm64.cnb",
          "stable_tag": "docker.io/heroku/buildpack-nodejs-npm-install:3.2.6_linux-arm64",
          "temporary_tag": "docker.io/heroku/buildpack-nodejs-npm-install:_9781380433_linux-arm64",
          "output_dir": "/home/runner/work/buildpacks-nodejs/buildpacks-nodejs/./packaged/aarch64-unknown-linux-musl/release/heroku_nodejs-npm-install"
        }
      ],
      "image_repository": "docker.io/heroku/buildpack-nodejs-npm-install",
      "stable_tag": "docker.io/heroku/buildpack-nodejs-npm-install:3.2.6",
      "temporary_tag": "docker.io/heroku/buildpack-nodejs-npm-install:_9781380433"
    }
    ```

When the automation runs successfully, either a single image or multiple images + a manifest list will have been published. Both the single image and manifest list are tagged with the value `matrix.stable_tag`. Therefore, if this tag exists, we shouldn't need to publish anything to Docker and we can skip that work.

Fixes #238

### DIFF
--- a/.github/workflows/_buildpacks-release.yml
+++ b/.github/workflows/_buildpacks-release.yml
@@ -147,7 +147,7 @@ jobs:
 
   publish-docker:
     name: Publish → Docker - ${{ matrix.buildpack_id }}
-    needs: [compile]
+    needs: [ compile ]
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
@@ -156,7 +156,7 @@ jobs:
     steps:
       # Composite buildpacks that depend on bash buildpacks (like
       # heroku/nodejs-function) refer to bash buildpacks by their source
-      # location rather than the packaged location. Other buildpacks don't
+      # location rather than the packaged location. Other buildpacks
       # don't need this step, so it's skipped where possible.
       - name: Checkout
         if: matrix.buildpack_type == 'composite'
@@ -186,7 +186,12 @@ jobs:
           username: ${{ secrets.docker_hub_user }}
           password: ${{ secrets.docker_hub_token }}
 
+      - name: Check if version is already on Docker Hub
+        id: check
+        run: echo "published_to_docker=$(docker manifest inspect "${{ matrix.stable_tag }}" &> /dev/null && echo 'true' || echo 'false')" >> $GITHUB_OUTPUT
+
       - name: Publish to temporary tags
+        if: steps.check.outputs.published_to_docker == 'false'
         env:
           TARGETS: ${{ toJSON(matrix.targets) }}
         run: |
@@ -222,7 +227,7 @@ jobs:
           fi
 
       - name: Promote temporary tags to stable tags
-        if: inputs.dry_run == false
+        if: inputs.dry_run == false && steps.check.outputs.published_to_docker == 'false'
         env:
           TARGETS: ${{ toJSON(matrix.targets) }}
         run: |
@@ -270,7 +275,7 @@ jobs:
 
   publish-github:
     name: Publish → GitHub Release
-    needs: [compile]
+    needs: [ compile ]
     runs-on: ${{ inputs.ip_allowlisted_runner }}
     steps:
       # Composite buildpacks that depend on bash buildpacks (like
@@ -332,7 +337,7 @@ jobs:
 
   publish-cnb-registry:
     name: Publish → CNB Registry - ${{ matrix.buildpack_id }}
-    needs: [compile, publish-docker]
+    needs: [ compile, publish-docker ]
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
@@ -367,7 +372,7 @@ jobs:
 
   update-builder:
     name: Update Builder
-    needs: [compile, publish-docker, publish-cnb-registry, publish-github]
+    needs: [ compile, publish-docker, publish-cnb-registry, publish-github ]
     runs-on: ${{ inputs.ip_allowlisted_runner }}
     steps:
       - name: Get token for GH application (Linguist)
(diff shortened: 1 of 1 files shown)


.github/workflows/_buildpacks-release.yml
```diff
@@ -147,7 +147,7 @@ jobs:
 
   publish-docker:
     name: Publish → Docker - ${{ matrix.buildpack_id }}
-    needs: [compile]
+    needs: [ compile ]
```
edmorley, line 150: Could you revert the whitespace formatting changes to the `needs` attributes?

The GitHub Actions workflow reference doesn't use whitespace padding style for these:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

And neither does https://www.yaml.info/learn/bestpractices.html nor Prettier in YAML mode.

edmorley (approved):
Thank you for fixing this! :-)

.github/workflows/_buildpacks-release.yml
colincasey: Hmm, must be a new formatting setting on RustRover. I don't remember my IDE adding those before. Anyways, I disabled that extra formatting and saved with the whitespace removed.

colincasey:
Verified it's working as expected in this test run:
https://github.com/heroku/buildpacks-nodejs/actions/runs/9799098390/job/27058768319

PR #289, opened 2025-03-11
Allow configuring the PR reviewer for cnb-builder-images PRs

edmorley:
Currently for CNB releases, the PR opened against `cnb-builder-images` doesn't have an explicit reviewer set by the automation, which means it uses that repo's `CODEOWNERS` default of requesting review from the whole Languages team.

Now, the workflow accepts an optional `reviewers` input, which can be used to select a specific person (eg the language owner) for review.

This will help reduce review-request-spam to other team members, and make it easier to see who each PR is waiting on in eg the Slack review reminders.

See:
https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#action-inputs
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs

GUS-W-18011095.

### DIFF
--- a/.github/workflows/_buildpacks-release.yml
+++ b/.github/workflows/_buildpacks-release.yml
@@ -22,6 +22,10 @@ on:
         type: boolean
         default: false
         description: Flag used for testing purposes to prevent actions that perform publishing operations from executing
+      reviewers:
+        description: A comma separated list of GitHub usernames from whom cnb-builder-images PR review will be requested, overriding the CODEOWNERS default of the Languages team.
+        type: string
+        required: false
       ip_allowlisted_runner:
         description: The GitHub Actions runner to use to run jobs that require IP allow-list privileges
         type: string
@@ -427,6 +431,7 @@ jobs:
             Update ${{ github.repository }} to v${{ needs.compile.outputs.version }}
 
             ${{ needs.compile.outputs.changelog }}
+          reviewers: ${{ inputs.reviewers }}
           path: ./cnb-builder-images
           branch: update/${{ github.repository }}
           delete-branch: true


.github/workflows/_buildpacks-release.yml
```diff
@@ -427,6 +431,7 @@ jobs:
             Update ${{ github.repository }} to v${{ needs.compile.outputs.version }}
 
             ${{ needs.compile.outputs.changelog }}
+          reviewers: ${{ inputs.reviewers }}
```
edmorley, line 434: The input will default to the empty string, which `peter-evans/create-pull-request` will handle fine:
https://github.com/peter-evans/create-pull-request/blob/58258f452b6c22066ed351398c99510be8470f5a/src/utils.ts#L16

Malax (approved):
Neat idea!

colincasey (approved):

edmorley:
Bah, so it looks like this might not have helped with reducing the email notification noise, since I still got an email requesting review from the languages team alias for a recent CNB release.

I believe this is due to `peter-evans/create-pull-request` setting the reviewer after opening the PR, by which point CODEOWNERS has already set the languages team as the reviewer.

My first thought was that perhaps the action could be updated to set the reviewer at the time of PR creation instead, however, checking the GitHub API docs it seems the create PR API doesn't support doing that - which is surprising given the GitHub UI for creating a PR does let you set those fields (but perhaps they use internal APIs).

One idea I had was to see whether opening the PR as draft first would help - since the CODEOWNERS reviewer doesn't get officially set until the PR is marked as ready for review, by which point the language owner would be set as a reviewer, and GitHub might skip the CODEOWNERS reviewer?

PR #296, opened 2025-04-02
Remove the `--locked` usage from the actions release step

edmorley:
Since this step, unlike all the others, actually has to update the lockfile for the version bump that's just occurred.

Fixes:
https://github.com/heroku/languages-github-actions/actions/runs/14218865177/job/39841765156#step:10:9

Follow-up to #293.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,7 +72,9 @@ jobs:
           cargo bump ${{ inputs.bump }}
 
       - name: Build
-        run: cargo build --release --locked
+        # Not using --locked since the lockfile will need updating for the version bump. The CI
+        # workflows for this repo will ensure the rest of the lockfile is up to date.
+        run: cargo build --release
 
       - name: Get release metadata
         id: metadata


.github/workflows/release.yml
```diff
@@ -72,7 +72,9 @@ jobs:
           cargo bump ${{ inputs.bump }}
 
       - name: Build
-        run: cargo build --release --locked
+        # Not using --locked since the lockfile will need updating for the version bump. The CI
+        # workflows for this repo will ensure the rest of the lockfile is up to date.
+        run: cargo build --release
```
colincasey, line 77: @edmorley could this potentially cause more than just the version to update if, for example, some changes were merged and then we forgot to run the release step for several days? 

If so, would an alternative be to:
- use `cargo fetch --locked` before the version bump
- perform the version bump step
- update the lockfile and produce the release build with `cargo build --offline --release`

edmorley: The CI run would fail on any change we make to `main` - there isn't really a way for the file to get out of sync.

Another option would be to remove the usage of `cargo bump` entirely, along with removing `version` from `Cargo.toml` - since they are not actually needed to create the binary and release (since we don't publish to crates.io) - `cargo bump` is instead really only used as a way to increment the semver version according to the major/minor/patch selection in the workflow input. We could then use a Bash (eg https://www.henryschmale.org/2019/04/30/incr-semver.html) or Rust approach to picking the new version string (which gets used in the Git tag etc).

colincasey (approved):

PR #357, opened 2026-05-08
Add reusable workflow for OIDC-based classic buildpack publishing

colincasey:
## Summary
- Adds `_classic-buildpack-publish.yml` reusable workflow that classic buildpack repos call on prepare-release PR merge
- Obtains GitHub Actions OIDC token and publishes to the buildpack registry via the new OIDC revisions endpoint
- Polls for completion, tags the commit, and creates a GitHub Release
- Input validation: registry URL allow-list, buildpack_id format check, version range sanity
- Concurrency control to prevent parallel publishes for the same buildpack
- Structured step summary output for all outcomes (success, CTC block, version mismatch, timeout, etc.)
- `qa` mode input for staging testing: publishes to staging registry, appends `-qa` to version tags, creates draft GitHub releases, and bypasses the default-branch guard

Depends on registry PRs:
- https://github.com/heroku/buildpack-registry/pull/527
- https://github.com/heroku/buildpack-registry/pull/528
- https://github.com/heroku/buildpack-registry/pull/533

W-22295383

### DIFF
--- a/.github/workflows/_classic-buildpack-publish.yml
+++ b/.github/workflows/_classic-buildpack-publish.yml
@@ -149,8 +149,9 @@ jobs:
 
           # Two-step so a corrupt repo / I/O error doesn't get silently treated as "tag absent":
           # `git show-ref --verify --quiet` exits 1 specifically for a missing ref and >1 for
-          # real errors. Once existence is confirmed, peeling to the commit cannot fail except
-          # on a real error, which we let propagate.
+          # real errors. Once existence is confirmed, resolving the tag with `^{commit}` gets
+          # the hash of the actual commit it points to instead of the tag object itself, and
+          # cannot fail except on a real error, which we let propagate.
           if git show-ref --tags --verify --quiet "refs/tags/${TAG}"; then
             tag_sha=$(git rev-parse --verify "refs/tags/${TAG}^{commit}")
           elif [[ $? -eq 1 ]]; then
--- a/.github/workflows/_classic-buildpack-publish.yml
+++ b/.github/workflows/_classic-buildpack-publish.yml
@@ -152,12 +152,17 @@ jobs:
           # real errors. Once existence is confirmed, resolving the tag with `^{commit}` gets
           # the hash of the actual commit it points to instead of the tag object itself, and
           # cannot fail except on a real error, which we let propagate.
-          if git show-ref --tags --verify --quiet "refs/tags/${TAG}"; then
+          #
+          # Capture the exit code once, then branch — using `$?` directly inside the elif/else
+          # arms would read the result of the preceding `[[ ]]` test, not git's exit code.
+          rc=0
+          git show-ref --tags --verify --quiet "refs/tags/${TAG}" || rc=$?
+          if [[ ${rc} -eq 0 ]]; then
             tag_sha=$(git rev-parse --verify "refs/tags/${TAG}^{commit}")
-          elif [[ $? -eq 1 ]]; then
+          elif [[ ${rc} -eq 1 ]]; then
             tag_sha=""
           else
-            echo "::error::git show-ref failed unexpectedly for refs/tags/${TAG} (exit $?)"
+            echo "::error::git show-ref failed unexpectedly for refs/tags/${TAG} (exit ${rc})"
             exit 1
           fi
 
--- a/.github/workflows/_classic-buildpack-publish.yml
+++ b/.github/workflows/_classic-buildpack-publish.yml
@@ -53,7 +53,7 @@ jobs:
       TAG_SUFFIX: ${{ inputs.qa && '-qa' || '' }}
     steps:
       - name: Checkout
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        uses: actions/checkout@v6
         with:
           fetch-depth: 0
           fetch-tags: true
--- a/.github/workflows/_classic-buildpack-publish.yml
+++ b/.github/workflows/_classic-buildpack-publish.yml
@@ -211,14 +211,24 @@ jobs:
           fi
 
           # Hit the API directly so we can branch on 200 vs 404 vs other rather than
-          # string-matching `gh release view` stderr.
+          # string-matching `gh release view` stderr. Capture stderr so transport-level
+          # failures (DNS, TLS, timeout) — which never produce an HTTP status line — can be
+          # surfaced in the diagnostic instead of being swallowed.
+          release_err_log=$(mktemp)
           release_status=$(gh api "repos/${GITHUB_REPOSITORY}/releases/tags/${TAG}" \
-            -i --silent 2>/dev/null | head -1 | awk '{print $2}') || true
+            -i --silent 2>"${release_err_log}" | head -1 | awk '{print $2}') || true
           case "${release_status}" in
             200) release_exists=true ;;
             404) release_exists=false ;;
             *)
               echo "::error::Unexpected status ${release_status:-unknown} from GitHub releases API for ${TAG}"
+              if [[ -s "${release_err_log}" ]]; then
+                {
+                  echo '```'
+                  cat "${release_err_log}"
+                  echo '```'
+                } >&2
+              fi
               exit 1
               ;;
           esac
--- a/.github/workflows/_classic-buildpack-publish.yml
+++ b/.github/workflows/_classic-buildpack-publish.yml
@@ -84,6 +84,14 @@ jobs:
             exit 1
           fi
 
+          # Validate buildpack_id format up-front (used by preflight, publish, and poll steps).
+          if [[ ! "${BUILDPACK_ID}" =~ ^[a-z0-9-]+/[a-z0-9-]+$ ]]; then
+            echo "::error::Invalid buildpack_id '${BUILDPACK_ID}' — expected format: owner/name (lowercase alphanumeric and hyphens)"
+            echo "## Publish failed (${MODE})" >> "${GITHUB_STEP_SUMMARY}"
+            echo "Invalid buildpack_id \`${BUILDPACK_ID}\` — expected format: \`owner/name\` (lowercase alphanumeric and hyphens only)." >> "${GITHUB_STEP_SUMMARY}"
+            exit 1
+          fi
+
       - name: Resolve tag and changes from changelog
         id: tag
         run: |
@@ -333,13 +341,6 @@ jobs:
             echo "::error::Untrusted registry_url '${REGISTRY_URL}' — refusing to send OIDC token"
             echo "## Publish failed (${MODE})" >> "${GITHUB_STEP_SUMMARY}"
             echo "Registry URL \`${REGISTRY_URL}\` is not in the allow-list. OIDC tokens will only be sent to trusted Heroku endpoints." >> "${GITHUB_STEP_SUMMARY}"
-            exit 1
-          fi
-
-          if [[ ! "${BUILDPACK_ID}" =~ ^[a-z0-9-]+/[a-z0-9-]+$ ]]; then
-            echo "::error::Invalid buildpack_id '${BUILDPACK_ID}' — expected format: owner/name (lowercase alphanumeric and hyphens)"
-            echo "## Publish failed (${MODE})" >> "${GITHUB_STEP_SUMMARY}"
-            echo "Invalid buildpack_id \`${BUILDPACK_ID}\` — expected format: \`owner/name\` (lowercase alphanumeric and hyphens only)." >> "${GITHUB_STEP_SUMMARY}"
             exit 1
           fi
 
--- a/.github/workflows/_classic-buildpack-publish.yml
+++ b/.github/workflows/_classic-buildpack-publish.yml
@@ -18,7 +18,8 @@ permissions:
   contents: write
 
 # Serialize publishes per buildpack so two rapid prepare-release merges (or a re-run racing a
-# fresh trigger) queue rather than racing the registry and the tag/release force-push.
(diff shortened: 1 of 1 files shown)


.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -0,0 +1,624 @@
...
+    runs-on: pub-hk-ubuntu-24.04-ip
+    # Single source of truth for the qa-vs-production routing — every step that needs the
+    # registry URL or tag suffix reads it from here, so the qa coupling can't drift across copies.
+    env:
+      MODE: ${{ inputs.qa && 'qa' || 'production' }}
+      REGISTRY_URL: ${{ inputs.qa && 'https://buildpack-registry-staging.herokuapp.com' || 'https://buildpack-registry.heroku.com' }}
+      TAG_SUFFIX: ${{ inputs.qa && '-qa' || '' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
edmorley: (We intentionally don't pin the official GitHub actions elsewhere at present - if we want to start doing so that should be it's own PR and change all uses at once)

```suggestion
        uses: actions/checkout@v6
```

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -0,0 +1,624 @@
...
+          # Release for this tag. The downstream steps mutate state in this order: tag →
+          # registry publish → release. Any partial-completion pattern (e.g. tag without a
+          # registry publish) is treated as divergence the buildpack author must clean up
+          # manually before re-running — the workflow never silently force-moves a tag,
+          # release, or published revision a customer may have pinned to.
+          encoded_buildpack_id=$(printf '%s' "${BUILDPACK_ID}" | jq -sRr @uri)
+
+          # Two-step so a corrupt repo / I/O error doesn't get silently treated as "tag absent":
+          # `git show-ref --verify --quiet` exits 1 specifically for a missing ref and >1 for
+          # real errors. Once existence is confirmed, peeling to the commit cannot fail except
```
edmorley: I don't understand what "peeling" means here? Is it a typo?

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -0,0 +1,624 @@
...
+          # on a real error, which we let propagate.
+          if git show-ref --tags --verify --quiet "refs/tags/${TAG}"; then
+            tag_sha=$(git rev-parse --verify "refs/tags/${TAG}^{commit}")
+          elif [[ $? -eq 1 ]]; then
+            tag_sha=""
+          else
+            echo "::error::git show-ref failed unexpectedly for refs/tags/${TAG} (exit $?)"
```
edmorley: Claude says:

`$?` in the `else` diagnostic reports the wrong exit code — line 157
```bash
if git show-ref --tags --verify --quiet "refs/tags/${TAG}"; then
  tag_sha=$(git rev-parse --verify "refs/tags/${TAG}^{commit}")
elif [[ $? -eq 1 ]]; then
  tag_sha=""
else
  echo "::error::git show-ref failed unexpectedly for refs/tags/${TAG} (exit $?)"
```
The `elif` branch correctly sees `git show-ref`'s exit code (under `set -e`, test commands in `if`/`elif` are exempt). But by the time the `else` arm runs, `[[ $? -eq 1 ]]` has already overwritten `$?` to `1`, so the diagnostic always prints `(exit 1)` regardless of the real `git` failure. **Fix:** capture `rc=$?` once after the `if`, then branch on `$rc`.

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -0,0 +1,624 @@
...
+          if [[ "${registry_published}" != "true" ]]; then
+            pending_revision_id=$(echo "${revisions}" | jq -r --arg tag "${TAG}" \
+              '[.[] | select(.ref == $tag and .status != "published" and .status != "failed")]
+               | sort_by(.created_at) | last | .id // empty')
+          fi
+
+          # Hit the API directly so we can branch on 200 vs 404 vs other rather than
+          # string-matching `gh release view` stderr.
+          release_status=$(gh api "repos/${GITHUB_REPOSITORY}/releases/tags/${TAG}" \
+            -i --silent 2>/dev/null | head -1 | awk '{print $2}') || true
```
edmorley: Should we avoid piping the stderr to dev null? It seems letting any errors be output to the logs might be preferable, otherwise errors that don't result in an HTTP status code, eg timeouts/network issues etc, won't display a useful error message.

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -0,0 +1,624 @@
...
+        run: |
+          # Use --notes-file rather than --notes so a multi-line body with backticks,
+          # quotes, and PR links round-trips verbatim — no shell-escaping risk.
+          notes_file=$(mktemp)
+          if [[ -n "${CHANGES//[[:space:]]/}" ]]; then
+            printf '%s\n' "${CHANGES}" > "${notes_file}"
+          else
+            # Fallback: heading matched but the section body was empty (or whitespace-only).
+            # A bare link beats publishing an empty release body.
+            echo "See [CHANGELOG.md](https://github.com/${{ github.repository }}/blob/main/CHANGELOG.md) for details." > "${notes_file}"
```
edmorley: Claude says:

Empty changelog body silently falls back — lines 119-131 + 601-607
`awk` extraction with no body produces `changes=""`, which is written to `$GITHUB_OUTPUT` without warning. The release-creation step has a hardcoded "See CHANGELOG.md" fallback — but it fires silently. An empty changelog section is almost always a formatting bug the author would want flagged. Add `::warning::CHANGELOG section for v${version} is empty` in the resolve step.

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -0,0 +1,624 @@
...
+            --request POST "${PUBLISH_URL}" \
+            --header "Accept: application/vnd.heroku+json; version=3.buildpack-registry" \
+            --header "Content-Type: application/json" \
+            --header "Authorization: Bearer ${OIDC_TOKEN}" \
+            --data "${REQUEST_BODY}") || true
+
+          http_code=$(echo "${response}" | tail -1 | tr -d '[:space:]')
+          body=$(echo "${response}" | sed '$d')
+
+          if [[ -z "${http_code}" ]]; then
```
edmorley, line 458: Claude says:

OIDC and publish-POST curl stderr discarded — lines 363-372 and 409-415
Connection failures (DNS, TLS handshake, post-retry transport) produce `HTTP unknown` and `No response from registry (connection failed after retries)` — both with the actual transport error thrown away. Capture stderr to a temp file and include in the step summary. Especially important for the publish POST since it's non-idempotent.

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -0,0 +1,624 @@
...
+
+          # Surface the misconfiguration directly rather than letting the curl below send empty
+          # headers and bottom out at the generic "Failed to obtain OIDC token" message.
+          : "${ACTIONS_ID_TOKEN_REQUEST_TOKEN:?id-token: write permission is required on the caller workflow}"
+          : "${ACTIONS_ID_TOKEN_REQUEST_URL:?id-token: write permission is required on the caller workflow}"
+
+          # Audience is intentionally the same for prod and staging — the registry validates
+          # the audience claim as a fixed string, and the per-environment routing is handled
+          # by REGISTRY_URL above. Changing the audience here would break the trust policy match.
+          # Default --retry transient set (5xx, 429, connection errors) — no --retry-all-errors,
```
edmorley: Should we be using `--retry-connrefused` too, to truly retry on connection errors as the comment says?

https://everything.curl.dev/usingcurl/downloads/retry.html#connection-refused

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -0,0 +1,624 @@
...
+            echo "Registry URL \`${REGISTRY_URL}\` is not in the allow-list. OIDC tokens will only be sent to trusted Heroku endpoints." >> "${GITHUB_STEP_SUMMARY}"
+            exit 1
+          fi
+
+          if [[ ! "${BUILDPACK_ID}" =~ ^[a-z0-9-]+/[a-z0-9-]+$ ]]; then
+            echo "::error::Invalid buildpack_id '${BUILDPACK_ID}' — expected format: owner/name (lowercase alphanumeric and hyphens)"
+            echo "## Publish failed (${MODE})" >> "${GITHUB_STEP_SUMMARY}"
+            echo "Invalid buildpack_id \`${BUILDPACK_ID}\` — expected format: \`owner/name\` (lowercase alphanumeric and hyphens only)." >> "${GITHUB_STEP_SUMMARY}"
+            exit 1
+          fi
```
edmorley: Should this BUILDPACK_ID check be in the earlier preflight checks step instead of here?

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -0,0 +1,624 @@
...
+            else
+              error_id="unknown"
+              error_msg="Empty or non-JSON response from registry (HTTP ${http_code})"
+            fi
+            case "${error_id}" in
+              ctc_denied)
+                override_url=$(echo "${body}" | jq -r '.override_url // "https://tps.heroku.tools/locks"')
+                if [[ ! "${override_url}" =~ ^https://(.*\.)?(heroku\.tools|heroku\.com)/ ]]; then
+                  override_url="https://tps.heroku.tools/locks"
+                fi
```
edmorley, line 540: Why is this fallback needed? Worth adding a comment?

edmorley (commented):

.github/workflows/_classic-buildpack-publish.yml
colincasey: Moved to the early **Resolve and validate publish configuration** step in 7fd4f8f — preflight, poll, and release-create all rely on BUILDPACK_ID being well-formed, so failing fast up-front makes the contract clearer.

.github/workflows/_classic-buildpack-publish.yml
colincasey: Reworded the comment in b545e20 to drop the "peeling" jargon — it now explains that `^{commit}` gets the hash of the commit the tag points to rather than the tag object itself.

.github/workflows/_classic-buildpack-publish.yml
colincasey: Good catch — fixed in 5eed3be. Capture `rc=$?` once after the show-ref call, then branch on `$rc`, so both the elif test and the error diagnostic see git's actual exit code.

.github/workflows/_classic-buildpack-publish.yml
colincasey: Fixed in 904bdc6. Stderr is now teed to a log file and dumped on the unexpected-status branch — so transport failures (DNS, TLS, timeout) that produce no HTTP status line surface in the diagnostic instead of being swallowed.

.github/workflows/_classic-buildpack-publish.yml
colincasey: Added in 64f4e81. The resolve step now emits a `::warning::` when the CHANGELOG section body is empty (whitespace-only counts as empty), so the formatting bug surfaces in the run summary before the release-creation fallback fires.

.github/workflows/_classic-buildpack-publish.yml
colincasey: Fixed in f369ecd. Both the OIDC token request and the publish POST now tee stderr to a temp file, and the failure summaries include curl's transport error in a fenced `Transport error:` block. Especially important for the publish POST since it's non-idempotent — the operator needs the underlying error to decide whether the request might still have reached the registry.

.github/workflows/_classic-buildpack-publish.yml
colincasey: Good point — added `--retry-connrefused` in d29149e for both the OIDC request and the publish POST, and reworded the comment to be accurate: bare `--retry` covers 5xx/429 and DNS/network errors but excludes ECONNREFUSED, so a refused connection (registry mid-deploy, for example) wasn't being retried. Safe for the non-idempotent publish POST since a refused connection means the request never reached the server, so a retry can't create a duplicate row.

.github/workflows/_classic-buildpack-publish.yml
colincasey: Added a comment in d98baa8. The fallback exists because we surface `override_url` as a clickable link in the step summary — a malformed or hostile value from the registry response could steer operators to an attacker page, so we restrict to the heroku.tools/heroku.com allow-list and fall back to the canonical TPS URL when the registry omits the field or returns something off-domain.

.github/workflows/_classic-buildpack-publish.yml
edmorley: @colincasey Shouldn't we fail though? The fallback ("See CHANGELOG.md") is kinda bad and IMO should never actually be used?

edmorley (approved):

.github/workflows/_classic-buildpack-publish.yml
colincasey: Agreed — fixed in 989e362. The resolve step now hard-fails (`::error::` + `exit 1`) when the CHANGELOG section body is empty, instead of warning. Dropped the unreachable "See CHANGELOG.md" fallback from the release-creation step now that `CHANGES` is guaranteed non-empty by the time we get there.

PR #358, opened 2026-05-14
Document auto-triggering release on prepare-release PR merge

### LINKED ISSUE #96: Automatically trigger the CNB release workflow when the "Prepare release" PR is merged
Currently the steps to perform a CNB release are:
1. Trigger the "Prepare release" workflow, and wait for it to open the "Prepare release PR"
2. Leave a review approval on the CNB release PR (the PR will then auto-merge)
3. Trigger a "Release buildpack" workflow targetting `main`, and wait for it to perform the release and eventually open the builder repo PR
4. Leave a review approval on the builder update PR (the PR will then auto-merge)

As a future optimisation, it seems we could skip step (3) by having the workflow auto-trigger when the "Prepare release" PR is merged.

The SF CLI has a workflow that does something similar:
https://github.com/salesforcecli/cli/blob/8aa0ab825cc4e7367690c0baa6011ee6134dc6b7/.github/workflows/create-github-release.yml#L4-L17

cc @colincasey

runesoerensen:
Adds `on: pull_request: [closed]` and an `if:` gate to the documented release workflow example, so consumers can opt into having the release workflow auto-trigger when the auto-generated "Prepare release" PR is merged. Manual `workflow_dispatch` is preserved for re-runs and dry-run testing.

README-only change; the shared `_buildpacks-release.yml` is unchanged. The gating logic lives in the consumer alongside the trigger, so each consumer's release config remains self-documenting.

Also see https://github.com/heroku/buildpacks-dotnet/pull/425

Resolves #96.

GUS-W-13871118

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ jobs:
     if: >-
       github.event_name == 'workflow_dispatch' ||
       (github.event.pull_request.merged == true &&
+       github.event.pull_request.head.repo.full_name == github.repository &&
        github.event.pull_request.head.ref == 'prepare-release')
     uses: heroku/languages-github-actions/.github/workflows/_buildpacks-release.yml@latest
     with:
--- a/README.md
+++ b/README.md
@@ -146,7 +146,8 @@ jobs:
       github.event_name == 'workflow_dispatch' ||
       (github.event.pull_request.merged == true &&
        github.event.pull_request.head.repo.full_name == github.repository &&
-       github.event.pull_request.head.ref == 'prepare-release')
+       github.event.pull_request.head.ref == 'prepare-release' &&
+       github.event.pull_request.user.login == 'heroku-linguist[bot]')
     uses: heroku/languages-github-actions/.github/workflows/_buildpacks-release.yml@latest
     with:
       app_id: ${{ vars.GH_APP_ID }}
--- a/README.md
+++ b/README.md
@@ -124,6 +124,13 @@ You can pin to:
 name: Release Buildpacks
 
 on:
+  # Auto-trigger when the "Prepare release" PR (created by the
+  # _buildpacks-prepare-release.yml workflow) is merged.
+  pull_request:
+    branches:
+      - main
+    types:
+      - closed
   workflow_dispatch:
 
 # Disable all GITHUB_TOKEN permissions, since the GitHub App token is used instead.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -124,6 +124,13 @@ You can pin to:
 name: Release Buildpacks
 
 on:
+  # Auto-trigger when the "Prepare release" PR (created by the
+  # _buildpacks-prepare-release.yml workflow) is merged.
+  pull_request:
+    branches:
+      - main
+    types:
+      - closed
```
edmorley, line 133: Should this be `merged` rather than `closed`?

README.md
```diff
@@ -132,6 +139,13 @@ permissions: {}
 jobs:
   release:
     name: Release
+    # On `pull_request`, only run for the merged auto-generated
+    # "Prepare release" PR (branch name set by
+    # _buildpacks-prepare-release.yml). Manual dispatches always run.
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      (github.event.pull_request.merged == true &&
+       github.event.pull_request.head.ref == 'prepare-release')
```
edmorley: Do we need additional filters like the username of the Linguist bot or some other attribute (eg PR title) here? I guess it's unlikely a human would accidentally use this branch name? And from a security/abuse point of view:
(a) running the release workflow is pretty safe to do at any point,
(b) only we can push to this branch

...though (b) raises the question - does this filter stop forked repo PRs using that branch name? Do we need a repo filter too? (Either that or we do add the linguist username filter after all)

README.md
runesoerensen: @edmorley I don't think `merged` is a valid [activity type](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request) for pull requests?

When the PR closed event fires it'll have `merged` set to `true`, so we're checking that in the conditional below (which seems to be [the recommended approach for this scenario](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges)?).

README.md
runesoerensen: Forks are already covered — GitHub [strips secrets and makes `GITHUB_TOKEN` read-only](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow) for any `pull_request` event from a fork (including `closed`), so the publishing-side surface is safe. But that doesn't stop someone with push access from pushing a `prepare-release` branch directly and merging it — and the cost of tightening is nothing. Pushed two commits:

- `head.repo.full_name == github.repository` — fails fast at `if:` evaluation, also documents intent.
- `user.login == 'heroku-linguist[bot]'` — ensures only the auto-generated PR can trigger a release.

README.md
edmorley: Ah I was looking at the REST API page I think. Ignore then :-)

edmorley (approved):

PR #360, opened 2026-05-14
Comment release progress on the prepare-release PR

### LINKED ISSUE #113: Provide guided notification as comments on the prepare PR
If we implement #96 then it should also be possible to provide some guided notifications as comments during the release workflow such as:
* A notification that the release process has started + a link to the running job
* A notification that the release process has finished + links to the published releases
* A notification providing a link to the PR for `heroku/builder` and details for what to do next
* A notification when the `heroku/builder` PR is merged (or closed) which completes the release process

These would be attached to the Prepare PR that initiates the process so that we could manage and see the state of the whole end-to-end process.

runesoerensen:
When the release workflow runs from a prepare-release PR merge (the auto-trigger path enabled by #358), post status comments back to that PR. Two on the happy path, one on failure:

* When the workflow starts (links to the run)
* When the `cnb-builder-images` PR is opened (links to that PR)
* When any job fails (links to the workflow run)

The aim is to give the team enough signal to act on what the auto-trigger surfaces: reassurance that a release started, a follow-up handle for the builder PR, and visibility when something goes wrong. Things that help folks outside the team track release status — Docker Hub tags, CNB registry entry, post-publish state — feel better addressed by enriching the GitHub release description in a follow-up.

Notifications are skipped on `workflow_dispatch` and dry runs, so behavior for manual / testing invocations is unchanged.

A fourth notification (\"builder PR merged → release complete\") was scoped out — it requires changes in `heroku/cnb-builder-images` and is better tracked separately.

Resolves #113.

GUS-W-13955895.

### DIFF
--- a/.github/workflows/_buildpacks-release.yml
+++ b/.github/workflows/_buildpacks-release.yml
@@ -386,7 +386,7 @@ jobs:
           fail_on_unmatched_files: true
 
       - name: Notify GitHub release published on prepare PR
-        if: github.event_name == 'pull_request' && inputs.dry_run == false
+        if: github.event_name == 'pull_request' && inputs.dry_run == false && steps.check.outputs.already_published_to_github == 'false'
         env:
           GH_TOKEN: ${{ steps.generate-token.outputs.token }}
           GH_REPO: ${{ github.repository }}
--- a/.github/workflows/_buildpacks-release.yml
+++ b/.github/workflows/_buildpacks-release.yml
@@ -73,6 +73,24 @@ jobs:
       version: ${{ steps.generate-buildpack-matrix.outputs.version }}
       changelog: ${{ steps.generate-changelog.outputs.changelog }}
     steps:
+      - name: Get token for prepare PR comment
+        if: github.event_name == 'pull_request'
+        uses: actions/create-github-app-token@v3
+        id: comment-token
+        with:
+          app-id: ${{ inputs.app_id }}
+          private-key: ${{ secrets.app_private_key }}
+
+      - name: Notify release started on prepare PR
+        if: github.event_name == 'pull_request'
+        env:
+          GH_TOKEN: ${{ steps.comment-token.outputs.token }}
+          GH_REPO: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: |
+          gh pr comment "$PR_NUMBER" --body "Release workflow started: $RUN_URL"
+
       - name: Checkout
         uses: actions/checkout@v6
         with:
@@ -367,6 +385,16 @@ jobs:
           files: "*.cnb"
           fail_on_unmatched_files: true
 
+      - name: Notify GitHub release published on prepare PR
+        if: github.event_name == 'pull_request' && inputs.dry_run == false
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+          GH_REPO: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          RELEASE_URL: ${{ github.server_url }}/${{ github.repository }}/releases/tag/v${{ needs.compile.outputs.version }}
+        run: |
+          gh pr comment "$PR_NUMBER" --body "GitHub release published: $RELEASE_URL"
+
   publish-cnb-registry:
     name: Publish → CNB Registry - ${{ matrix.buildpack_id }}
     needs: [compile, publish-docker]
@@ -469,3 +497,21 @@ jobs:
         run: gh pr merge --auto --squash --repo heroku/cnb-builder-images "${{ steps.pr.outputs.pull-request-number }}"
         env:
           GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+
+      - name: Get token for prepare PR comment
+        if: github.event_name == 'pull_request' && inputs.dry_run == false && steps.pr.outputs.pull-request-operation == 'created'
+        uses: actions/create-github-app-token@v3
+        id: comment-token
+        with:
+          app-id: ${{ inputs.app_id }}
+          private-key: ${{ secrets.app_private_key }}
+
+      - name: Notify builder PR opened on prepare PR
+        if: github.event_name == 'pull_request' && inputs.dry_run == false && steps.pr.outputs.pull-request-operation == 'created'
+        env:
+          GH_TOKEN: ${{ steps.comment-token.outputs.token }}
+          GH_REPO: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          BUILDER_PR_URL: ${{ steps.pr.outputs.pull-request-url }}
+        run: |
+          gh pr comment "$PR_NUMBER" --body "Builder PR opened: $BUILDER_PR_URL — review and merge to complete the release."
--- a/.github/workflows/_buildpacks-release.yml
+++ b/.github/workflows/_buildpacks-release.yml
@@ -385,16 +385,6 @@ jobs:
           files: "*.cnb"
           fail_on_unmatched_files: true
 
-      - name: Notify GitHub release published on prepare PR
-        if: github.event_name == 'pull_request' && inputs.dry_run == false && steps.check.outputs.already_published_to_github == 'false'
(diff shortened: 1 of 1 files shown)


.github/workflows/_buildpacks-release.yml
```diff
@@ -367,6 +384,15 @@ jobs:
...
           fail_on_unmatched_files: true
 
+      - name: Notify GitHub release published on prepare PR
+        if: github.event_name == 'pull_request' && inputs.dry_run == false
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+          GH_REPO: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          RELEASE_URL: ${{ github.server_url }}/${{ github.repository }}/releases/tag/v${{ needs.compile.outputs.version }}
+        run: gh pr comment "$PR_NUMBER" --body "GitHub release published: $RELEASE_URL"
```
edmorley: A Claude code review of the PR spotted this:

1. **"GitHub release published" notification fires on idempotent re-runs even when nothing was published** (`_buildpacks-release.yml:387-394`)
   The notification's gate is:
   ```yaml
   if: github.event_name == 'pull_request' && inputs.dry_run == false
   ```
   But the create-release step right above it has an additional check:
   ```yaml
   if: inputs.dry_run == false && steps.check.outputs.already_published_to_github == 'false'
   ```
   So if `steps.check.outputs.already_published_to_github == 'true'` (release already exists from a prior run), the create-release step is skipped, but the notification *still fires*, posting "GitHub release published: <url>" even though nothing was published this run.

   Compare with the "Notify builder PR opened" step (lines 506+), which correctly gates on `steps.pr.outputs.pull-request-operation == 'created'` — it only fires when something actually happened. The GitHub-release notification should mirror that and add `&& steps.check.outputs.already_published_to_github == 'false'`. Either that or rephrase the comment ("GitHub release available: …") so the message is accurate when the release pre-existed.

.github/workflows/_buildpacks-release.yml
```diff
@@ -73,6 +73,23 @@ jobs:
       version: ${{ steps.generate-buildpack-matrix.outputs.version }}
       changelog: ${{ steps.generate-changelog.outputs.changelog }}
     steps:
+      - name: Get token for prepare PR comment
+        if: github.event_name == 'pull_request'
```
edmorley, line 77: Do we need to filter on event name? I think the comments would still be helpful for manual dispatch? Otherwise anyone just reading the PR comments might get confused as to current state?

edmorley (commented):

Zooming out for a second, I think the reasons we want these comments are:
1. To raise awareness if the release (or a part of it) didn't succeed (particularly after #361), so we know to take a look/retrigger/debug etc.
2. To reassure us that the automatic releases after prepare PR merge are indeed happening (until we gain confidence with the approach in #358)
3. To help those outside the language team see what the status of a release is, and to "follow along" etc.
4. To remind us that we do need to merge the builder PR too (though I guess longer term ideally that part might be automatic too).
5. To reduce the number of clicks for us to find one of the release outputs (eg the actions workflow logs, the GitHub release, or the builder PR), by having them pasted directly in the comments.

Of all those, I think (1) is probably the most important, however, as-is the implementation in this PR doesn't leave any comments on job failure.

Also, on the happy path, the implementation currently leaves lots of small comments on the PR, rather than one summary. I guess the advantage of that is if one step fails you get to see the incremental comments, though it does cause extra email notification noise.

Another approach might be to:
1. Keep the initial "Release workflow triggered" message on the PR (since it's useful to know it's started)
2. However, update the per-step fine-grained status update messages so they only report on failure (so we get error messages early)
3. Then add a final new step that depends on all the others, that gives a summary of all the things (along with the "remember to review the builder PR"). This summary could either be run only on success, or perhaps could be shown for both success and failure (with different contents)

Or perhaps even simpler, we could skip (2), and just have the final summary? After all, a failing GHA workflow can't be retriggered until the whole workflow has finished, so even if we post a comment about a failure in one step early, a human can't do much about it until the other jobs have finished anyway?

.github/workflows/_buildpacks-release.yml
runesoerensen: Good point. The trouble is that `github.event.pull_request` is `null` on `workflow_dispatch`, so the workflow doesn't have an obvious way to know which PR to comment on. We could maybe try to detect it from the git history (eg the most recent merged `prepare-release` PR), though that feels a bit brittle.

A simpler option might be a `prepare_pr_number` input as a follow-up, defaulting to `github.event.pull_request.number` for the auto-trigger path.

Alternatively I suppose we could drop manual dispatch entirely, and only allow releases via prepare-release merge? "Re-run failed jobs" would handle most recovery scenarios since it reuses the original event context, so the main thing we'd lose is the dry-run feature.

That said, I'm not sure it's worth it in the context of letting folks use the "Prepare Release" PR comments to track current state... If we want to lets folks outside the team track the state of a release, I wonder if it'd be better to update the GitHub release description with relevant information when the workflow succeeds (which could then include links to the builder PR, docker hub/buildpack registry references etc)?

.github/workflows/_buildpacks-release.yml
runesoerensen: Fixed in ac68b6d (I suspect we may end up dropping these per-step notifications entirely in favor of the [summary approach you suggested](https://github.com/heroku/languages-github-actions/pull/360#pullrequestreview-4289963642), but worth fixing regardless :))

runesoerensen:
Thanks for the detailed review — really helpful pushback.

Walked through your 5 goals to see which the existing comments were actually serving:

1. **Failure visibility** — the only signal we don't already have today, and especially important after #361 since `update-builder` running in parallel means we can end up in a partial-success state that's otherwise silent
2. **Reassurance that auto-releases are happening** — the "Release workflow started" comment covers this
3. **Help non-team people see release status** — I don't think they really are (or should be) watching comments on PRs for this? As suggested in https://github.com/heroku/languages-github-actions/pull/360#discussion_r3256229843, growing the GitHub release description into a fuller landing page (with links to the builder PR, Docker Hub tags, CNB registry entry) feels like a more viable approach if we want to surface this more broadly
4. **Builder PR merge reminder** — the existing "Builder PR opened" comment covers this; the standard review-request notification covers the rest
5. **Reduce clicks to find release outputs** — covered by the existing "Release workflow started" and "Builder PR opened" comments for what the team needs in-flight; the release-description angle from (3) could cover the rest

Updated the PR to:

* Keep the "Release workflow started" comment (goals 2, 5)
* Keep the "Builder PR opened" comment (goals 4, 5)
* Drop the "GitHub release published" comment — the release page surfaces this anyway, and the failure notification below catches the case where publishing didn't happen
* Add a `notify-failure` job that fires only on failure, posting "Release workflow finished with failures — see the workflow run at <url>" (goal 1)

The framing I landed on: the comments that stay are the ones that help *the team* act (start + builder PR for follow-up + failure for retry/debug). The release page is arguably a better home for things that help folks *outside the team* track release status, but that's a separate direction worth weighing on its own merits as a follow-up rather than committing to here.

That gives us two comments per release on the happy path (start + builder PR) and a third on failure. The failure trigger is \`if: failure() && github.event_name == 'pull_request'\` which captures any non-success state without us needing to enumerate jobs in the body.

On the email notification noise you raised: the prepare PR's subscribers would be emailed for each comment, so 2 per release on the happy path and 1–3 on failure depending on how far the workflow got. Who that ends up reaching depends on the prepare PR's review state — for buildpacks where the prepare PR was reviewed by an individual (e.g. heroku/buildpacks-dotnet#422) it's just the approver, but for buildpacks where \`@heroku/languages\` is also requested via CODEOWNERS (e.g. heroku/buildpacks-php#309) the team is notified at request time per [the team-review docs](https://docs.github.com/en/organizations/organizing-members-into-teams/managing-code-review-settings-for-your-team) — whether team members continue receiving comment-level notifications post-merge depends on individual notification settings, which I haven't been able to nail down precisely. Worth flagging — though it's tighter than the original implementation in both cases.

Let me know what you think.

.github/workflows/_buildpacks-release.yml
edmorley: Ah if the event payload doesn't have the required info then yeah let's not worry about this. I think it's useful to keep manual dispatch as an emergency fallback option rather than removing it. (And I suspect we won't need to be using it much anyway)

edmorley (approved):

PR #361, opened 2026-05-14
Trim unnecessary dependencies from update-builder job

### LINKED ISSUE #117: Remove unnecessary dependencies from the "Update builder" job
Currently the "Update the builder" job runs at the very end, since it's declared as having a dependency on all of the other jobs:

```mermaid
flowchart LR
    A[Compile buildpacks] --> B[Publish to Docker Hub]
    B --> C[Publish to CNB registry]
    A --> D[Create GitHub release]
    C & D --> E[Update builder]
```

However, opening the builder PR only actually needs the "Publish to Docker Hub" step to have been completed. ie:

```mermaid
flowchart LR
    A[Compile buildpacks] --> B[Publish to Docker Hub]
    B --> C[Publish to CNB registry]
    A --> D[Create GitHub release]
    B --> E[Update builder]
```

By removing these unnecessary dependencies from the workflow definition, we:
1. Ensure that we can still publish builder updates if the CNB registry is having downtime (xref https://salesforce.quip.com/QRwnAaGborAF#temp:C:RCc6106ab8f1a3e4f37968a5b44a)
2. Speed up the end to end time of publishing a CNB, since more of the jobs can now run in parallel

However, we might want to wait to do this until we've done #113, since currently the only signal buildpack maintainers have for failed jobs is that the builder update PR never gets opened - so we want to make sure there is still a way to see that eg the CNB registry job has failed.

GUS-W-13956414.

runesoerensen:
`update-builder` only consumes the published Docker Hub image digest (read via `crane digest`) plus `compile` outputs — it has no functional dependency on `publish-cnb-registry` or `publish-github`. The current over-specified `needs:` list serializes the workflow unnecessarily and couples the cnb-builder-images PR to CNB registry availability; when `registry.buildpacks.io` is degraded, builder updates are blocked for no good reason.

Trims `needs:` to `[compile, publish-docker]` so the builder PR can open in parallel with `publish-cnb-registry` and `publish-github`. Speeds up end-to-end release time and decouples builder updates from CNB registry uptime.

Issue #117 noted this should wait until per-step failure signal exists, since previously "the builder PR never opened" was the only indication something went wrong. That prerequisite is met by #360, which posts comments to the prepare PR for each release step.

Resolves #117.

GUS-W-13956414

### DIFF
--- a/.github/workflows/_buildpacks-release.yml
+++ b/.github/workflows/_buildpacks-release.yml
@@ -427,7 +427,12 @@ jobs:
 
   update-builder:
     name: Update Builder
-    needs: [compile, publish-docker, publish-cnb-registry, publish-github]
+    # `update-builder` only needs the published Docker Hub image (read via
+    # `crane digest`) plus `compile` outputs. Keeping it independent of
+    # `publish-cnb-registry` and `publish-github` lets the builder PR open
+    # in parallel with those jobs and means CNB-registry downtime no
+    # longer blocks builder updates.
+    needs: [compile, publish-docker]
     runs-on: ${{ inputs.ip_allowlisted_runner }}
     steps:
       - name: Get token for GH application (Linguist)


edmorley (approved):

.github/workflows/_buildpacks-release.yml
```diff
@@ -427,7 +427,12 @@ jobs:
 
   update-builder:
     name: Update Builder
-    needs: [compile, publish-docker, publish-cnb-registry, publish-github]
+    # `update-builder` only needs the published Docker Hub image (read via
+    # `crane digest`) plus `compile` outputs. Keeping it independent of
+    # `publish-cnb-registry` and `publish-github` lets the builder PR open
+    # in parallel with those jobs and means CNB-registry downtime no
+    # longer blocks builder updates.
+    needs: [compile, publish-docker]
```
edmorley, line 435: GitHub renders this really unhelpfully in the workflow summary - makes it look like the update-builder step has a dependency on the publish-github step (when it doesn't any more) - which made me come back to double check the `needs` was correct 😅 

Image

(Not that we need to/can do anything about this; just mentioning as an observation)

PR #369, opened 2026-06-04
Add Honeycomb release markers to classic buildpack publish

colincasey:
Adds an optional `honeycomb_api_key` secret to the classic buildpack publish workflow. When a caller passes it, a best-effort, idempotent marker is posted to the Honeycomb `builds` dataset on a successful **production** publish.

## What

- New optional `honeycomb_api_key` secret on `_classic-buildpack-publish.yml`'s `workflow_call` interface.
- A new **Post Honeycomb release marker** step appended after *Create GitHub Release*.

## Behavior

- **Opt-in / no-op:** skipped entirely when the secret isn't passed, so other language teams are unaffected.
- **Production-only:** gated on `env.MODE == 'production'` — QA/staging publishes don't post markers.
- **Idempotent (create-or-update):** dedups on the release URL, so re-running an idempotent publish updates the existing marker instead of stacking duplicates — matching the rest of this workflow.
- **Best-effort:** every API call ends in `|| echo "::warning::..."` and never exits non-zero; a Honeycomb outage can't fail an already-successful publish.

The dataset slug `builds` is hardcoded (shared by all classic buildpacks, gated by the API key, not sensitive). Only the API key is a secret.

W-22846373

### DIFF
--- a/.github/workflows/_classic-buildpack-publish.yml
+++ b/.github/workflows/_classic-buildpack-publish.yml
@@ -25,6 +25,12 @@ on:
           Slack Workflow Builder webhook URL for #heroku-languages-ops. A message is posted
           here if a production publish fails.
         required: true
+      honeycomb_api_key:
+        description: >-
+          Honeycomb env API key with the 'create markers' scope. When set, a release
+          marker is posted to the `builds` dataset on a successful production publish.
+          Omit to disable (no-op for callers that don't opt in).
+        required: false
 
 # Full set of permissions the CALLING workflow must grant — a reusable workflow's GITHUB_TOKEN
 # is capped by what the caller provides. Each job below narrows to only what it needs via its
--- a/.github/workflows/_classic-buildpack-publish.yml
+++ b/.github/workflows/_classic-buildpack-publish.yml
@@ -758,6 +758,39 @@ jobs:
             exit 1
           fi
 
+      - name: Post Honeycomb release marker
+        # No-op when the secret is absent (other teams) or in QA mode. Secrets can't be
+        # referenced in `if:`, so the key is mirrored to env and tested there.
+        if: env.HONEYCOMB_API_KEY != '' && env.MODE == 'production'
+        env:
+          HONEYCOMB_API_KEY: ${{ secrets.honeycomb_api_key }}
+          HONEYCOMB_DATASET: builds          # shared dataset for all classic buildpacks
+          BUILDPACK_ID: ${{ inputs.buildpack_id }}
+          TAG: ${{ steps.tag.outputs.name }}
+        run: |
+          api="https://api.honeycomb.io/1/markers/${HONEYCOMB_DATASET}"
+          url="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/releases/tag/${TAG}"
+          msg="${BUILDPACK_ID} ${TAG}"
+          body=$(jq -n --arg msg "$msg" --arg url "$url" '{message: $msg, type: "deploy", url: $url}')
+          auth=(-H "X-Honeycomb-Team: ${HONEYCOMB_API_KEY}" -H "Content-Type: application/json")
+
+          # Dedup on the release URL: stable identity for this release. A re-run of an idempotent
+          # publish updates the existing marker rather than stacking a duplicate line on the board.
+          # Every path below warns on failure and never exits non-zero — a Honeycomb outage must
+          # not fail an already-successful publish.
+          existing=$(curl -sS --retry 3 --retry-all-errors --max-time 30 "${auth[@]}" "${api}" \
+            | jq -r --arg url "$url" 'map(select(.url == $url)) | first | .id // empty' 2>/dev/null) || existing=""
+
+          if [[ -n "$existing" ]]; then
+            curl -sS --retry 3 --retry-all-errors --max-time 30 -X PUT "${api}/${existing}" "${auth[@]}" -d "${body}" \
+              >/dev/null && echo "Updated existing release marker (${existing})" \
+              || echo "::warning::Failed to update Honeycomb release marker (publish succeeded regardless)"
+          else
+            curl -sS --retry 3 --retry-all-errors --max-time 30 -X POST "${api}" "${auth[@]}" -d "${body}" \
+              >/dev/null && echo "Created release marker for ${TAG}" \
+              || echo "::warning::Failed to create Honeycomb release marker (publish succeeded regardless)"
+          fi
+
   notify-failure:
     name: Notify Release Failure
     needs: [publish]
--- a/.github/workflows/_classic-buildpack-publish.yml
+++ b/.github/workflows/_classic-buildpack-publish.yml
@@ -775,18 +775,17 @@ jobs:
           auth=(-H "X-Honeycomb-Team: ${HONEYCOMB_API_KEY}" -H "Content-Type: application/json")
 
           # Dedup on the release URL: stable identity for this release. A re-run of an idempotent
-          # publish updates the existing marker rather than stacking a duplicate line on the board.
-          # Every path below warns on failure and never exits non-zero — a Honeycomb outage must
-          # not fail an already-successful publish.
-          existing=$(curl -sS --retry 3 --retry-all-errors --max-time 30 "${auth[@]}" "${api}" \
-            | jq -r --arg url "$url" 'map(select(.url == $url)) | first | .id // empty' 2>/dev/null) || existing=""
+          # publish finds the existing marker and skips, rather than stacking a duplicate line on
+          # the board. The marker content is fully derived from the tag, so there's nothing to
+          # update. This warns on failure and never exits non-zero — a Honeycomb outage must not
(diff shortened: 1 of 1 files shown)


.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -686,3 +693,36 @@ jobs:
...
+          auth=(-H "X-Honeycomb-Team: ${HONEYCOMB_API_KEY}" -H "Content-Type: application/json")
+
+          # Dedup on the release URL: stable identity for this release. A re-run of an idempotent
+          # publish updates the existing marker rather than stacking a duplicate line on the board.
+          # Every path below warns on failure and never exits non-zero — a Honeycomb outage must
+          # not fail an already-successful publish.
+          existing=$(curl -sS --retry 3 --retry-all-errors --max-time 30 "${auth[@]}" "${api}" \
+            | jq -r --arg url "$url" 'map(select(.url == $url)) | first | .id // empty' 2>/dev/null) || existing=""
+
+          if [[ -n "$existing" ]]; then
```
edmorley, line 790: Do we want to be updating the entry if an existing one exists? It feels like we can/should just skip in that case?

In which case this also means the implementation here can be written with much less duplication.

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -686,3 +693,36 @@ jobs:
...
+            curl -sS --retry 3 --retry-all-errors --max-time 30 -X PUT "${api}/${existing}" "${auth[@]}" -d "${body}" \
+              >/dev/null && echo "Updated existing release marker (${existing})" \
+              || echo "::warning::Failed to update Honeycomb release marker (publish succeeded regardless)"
+          else
+            curl -sS --retry 3 --retry-all-errors --max-time 30 -X POST "${api}" "${auth[@]}" -d "${body}" \
```
edmorley: Could you add the missing standard curl best practices like `--fail` and `--connect-timeout`? (Without `--fail`, we won't get the annotation when the API returns a non-200 value etc)

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -686,3 +693,36 @@ jobs:
...
+          url="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/releases/tag/${TAG}"
+          msg="${BUILDPACK_ID} ${TAG}"
+          body=$(jq -n --arg msg "$msg" --arg url "$url" '{message: $msg, type: "deploy", url: $url}')
+          auth=(-H "X-Honeycomb-Team: ${HONEYCOMB_API_KEY}" -H "Content-Type: application/json")
+
+          # Dedup on the release URL: stable identity for this release. A re-run of an idempotent
+          # publish updates the existing marker rather than stacking a duplicate line on the board.
+          # Every path below warns on failure and never exits non-zero — a Honeycomb outage must
+          # not fail an already-successful publish.
+          existing=$(curl -sS --retry 3 --retry-all-errors --max-time 30 "${auth[@]}" "${api}" \
```
edmorley: I don't think we should pipe errors to `/dev/null` - let them be printed as is (only the stdout gets captured by the `$()` so the stderr is fine to still print, since it will be shown directly in the logs)

edmorley (dismissed):
Sorry missed these on the first pass, only realised when looking at the similar https://github.com/heroku/cnb-builder-images/pull/989

.github/workflows/_classic-buildpack-publish.yml
colincasey: Fixed in d910c1f

.github/workflows/_classic-buildpack-publish.yml
colincasey: Fixed in d910c1f

.github/workflows/_classic-buildpack-publish.yml
colincasey: Fixed in d910c1f

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -752,6 +758,38 @@ jobs:
...
+          body=$(jq -n --arg msg "$msg" --arg url "$url" '{message: $msg, type: "deploy", url: $url}')
+          auth=(-H "X-Honeycomb-Team: ${HONEYCOMB_API_KEY}" -H "Content-Type: application/json")
+
+          # Dedup on the release URL: stable identity for this release. A re-run of an idempotent
+          # publish finds the existing marker and skips, rather than stacking a duplicate line on
+          # the board. The marker content is fully derived from the tag, so there's nothing to
+          # update. This warns on failure and never exits non-zero — a Honeycomb outage must not
+          # fail an already-successful publish.
+          existing=$(curl -sS --fail --connect-timeout 10 --retry 3 --retry-all-errors --max-time 30 "${auth[@]}" "${api}" \
+            | jq -r --arg url "$url" 'map(select(.url == $url)) | first | .id // empty') || existing=""
```
edmorley: Similar to https://github.com/heroku/cnb-builder-images/pull/989#discussion_r3437751040 I think we should remove all of these "silently ignore failure" cases. Since this job is the last in the workflow, it's fine if it fails - and will make it easier for us to spot issues (and know to retrigger it) etc.

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -752,6 +758,38 @@ jobs:
...
+
+          if [[ -n "$existing" ]]; then
+            echo "Release marker already exists (${existing}) — skipping"
+          else
+            curl -sS --fail --connect-timeout 10 --retry 3 --retry-all-errors --max-time 30 -X POST "${api}" "${auth[@]}" -d "${body}" \
+              >/dev/null && echo "Created release marker for ${TAG}" \
```
edmorley: Same as https://github.com/heroku/cnb-builder-images/pull/989#discussion_r3437758822 for `/dev/null`

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -752,6 +758,38 @@ jobs:
...
+        env:
+          HONEYCOMB_API_KEY: ${{ secrets.honeycomb_api_key }}
+          HONEYCOMB_DATASET: builds          # shared dataset for all classic buildpacks
+          BUILDPACK_ID: ${{ inputs.buildpack_id }}
+          TAG: ${{ steps.tag.outputs.name }}
+        run: |
+          api="https://api.honeycomb.io/1/markers/${HONEYCOMB_DATASET}"
+          url="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/releases/tag/${TAG}"
+          msg="${BUILDPACK_ID} ${TAG}"
+          body=$(jq -n --arg msg "$msg" --arg url "$url" '{message: $msg, type: "deploy", url: $url}')
```
edmorley: Similar to https://github.com/heroku/cnb-builder-images/pull/989#discussion_r3437835933 we should probably make the marker `type` consistent here compared to the CNB builder markers etc.

edmorley: Similar to https://github.com/heroku/cnb-builder-images/pull/989/changes#r3437849627 should we be setting `start_time`?

edmorley (commented):
Thank you for the updates!

I've left some more comments - in addition to those, could you also cross-check this PR with https://github.com/heroku/cnb-builder-images/pull/989 in general and make sure the two are consistent.

I'm now away from tomorrow until Wed - so happy for someone else on the team to re-review once those changes are made, to unblock you :-)

.github/workflows/_classic-buildpack-publish.yml
colincasey: Done in 9a690fe — dropped the `::warning::` fallback. With `--fail` and bash's default `pipefail`, a Honeycomb error now fails this (last) step red so we know to retrigger, without affecting the already-published release.

.github/workflows/_classic-buildpack-publish.yml
colincasey: Done in 9a690fe — removed the `>/dev/null` redirect.

.github/workflows/_classic-buildpack-publish.yml
colincasey: Both are already `deploy` (#369 here and #989 after your alignment there), so no change needed on this one.

colincasey: Added in 9a690fe — set from the release commit timestamp (`git log -1 --format=%ct`) rather than "now", so a delayed/re-triggered run still anchors the marker to the release. Matches the approach in #989.

colincasey:
Cross-checked against [#989](https://github.com/heroku/cnb-builder-images/pull/989) — the marker step now mirrors it: shared `curl` array, fail-loud on error, dedup-and-skip on existing marker, and `start_time` from the commit timestamp. Both use `type: deploy`. The one intentional difference is the dataset (`builds` for classic vs `heroku-builds-production` for CNBs).

edmorley (approved):
Thank you for the updates + doing this in general! :-)

PR #370, opened 2026-06-05
Add publish notifications to the classic buildpack release workflow

colincasey:
## Summary

Surfaces classic-buildpack publish progress and outcome outside the Actions run, so a failed publish can't go unnoticed. Brings the classic pipeline to parity with the CNB `_buildpacks-release.yml` `notify-start`/`notify-failure` pattern.

Adds two sibling jobs to the shared `_classic-buildpack-publish.yml`:

- **`notify-start`** — comments `Release workflow started: <run-url>` on the triggering "Prepare release" PR. Gated to merged production PRs (`pull_request` + `merged == true` + `!qa`), mirroring the publish job's own precondition.
- **`notify-failure`** (`needs: [publish]`, `if: failure()`) — comments `Release workflow finished with failures …` on the PR, **and** posts to `#heroku-languages-ops` via a Slack Workflow Builder webhook. Both notifications fire for **production prepare-PR failures only** — manual `workflow_dispatch` runs are for testing and stay silent.

Supporting changes:

- PR comments and the GitHub Release are now authored by the **Linguist GitHub App** (minted per-job from `app_id`/`app_private_key`), rather than the generic `github-actions` user — matching the CNB automation. Both notify jobs run on the IP-allowlisted runner (required for the App token mint) with `permissions: {}`.
- The version tag push stays on `github.token` (`contents: write`): a lightweight tag carries no identity, and pushing it via the App token would risk triggering downstream `on: push: tags` workflows.
- **Permissions restructure** — the workflow-level block now grants only `id-token` + `contents` (the union the `publish` job needs); the notify jobs need no `GITHUB_TOKEN` scopes.

## Caller changes (separate, not in this PR)

To adopt this, a classic buildpack's `release.yml` must pass three new required parameters: the `app_id` input, the `app_private_key` secret, and the `slack_webhook_url` secret. No `pull-requests: write` grant is needed (the notify jobs author comments via the App token). The `heroku-buildpack-nodejs` caller update + Slack webhook provisioning are tracked separately and must be coordinated with this change.

[W-22900983](https://gus.lightning.force.com/a07EE00002bkd3nYAA)

### DIFF
--- a/.github/workflows/_classic-buildpack-publish.yml
+++ b/.github/workflows/_classic-buildpack-publish.yml
@@ -12,22 +12,29 @@ on:
         type: boolean
         required: false
         default: false
+      app_id:
+        description: Application ID of the GitHub application (e.g. the Linguist App) used to comment on the prepare PR and author the GitHub Release.
+        type: string
+        required: true
     secrets:
-      SLACK_WEBHOOK_URL:
+      app_private_key:
+        description: Private key of the GitHub application (e.g. the Linguist App).
+        required: true
+      slack_webhook_url:
         description: >-
-          Slack Workflow Builder webhook URL for #heroku-languages-ops. When set, a message
-          is posted if a production publish fails. Omit to disable (no-op for callers that
-          don't opt in).
-        required: false
+          Slack Workflow Builder webhook URL for #heroku-languages-ops. A message is posted
+          here if a production publish fails.
+        required: true
 
 # Full set of permissions the CALLING workflow must grant — a reusable workflow's GITHUB_TOKEN
 # is capped by what the caller provides. Each job below narrows to only what it needs via its
-# own `permissions:` block (the actual enforcement): `publish` never carries `pull-requests`,
-# and the notify jobs never carry `id-token`/`contents`.
+# own `permissions:` block (the actual enforcement). PR comments and the GitHub Release are
+# authored by the Linguist GitHub App token (minted per-job from `app_id`/`app_private_key`),
+# matching the CNB release automation — so the notify jobs need no GITHUB_TOKEN scopes at all,
+# and `publish` only needs `id-token` (OIDC mint) plus `contents` (push the version tag).
 permissions:
   id-token: write
   contents: write
-  pull-requests: write
 
 # Serialize publishes per buildpack so two rapid prepare-release merges (or a re-run racing a
 # fresh trigger) queue rather than racing the pre-flight check and the tag/registry/release
@@ -55,22 +62,27 @@ jobs:
       github.event_name == 'pull_request'
       && github.event.pull_request.merged == true
       && !inputs.qa
-    runs-on: ubuntu-24.04
-    permissions:
-      pull-requests: write     # comment on the triggering prepare-release PR
+    # Runs on the IP-allowlisted runner because the Linguist App token mint hits the heroku
+    # org's IP allow list, which GitHub-hosted runners aren't on. The comment is authored by
+    # the App token (not github.token), so the job needs no GITHUB_TOKEN scopes of its own.
+    runs-on: pub-hk-ubuntu-24.04-ip
+    permissions: {}
     steps:
+      - name: Get token for prepare PR comment
+        uses: actions/create-github-app-token@v3
+        id: comment-token
+        with:
+          app-id: ${{ inputs.app_id }}
+          private-key: ${{ secrets.app_private_key }}
+
       - name: Comment on prepare PR
         env:
-          GH_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ steps.comment-token.outputs.token }}
           GH_REPO: ${{ github.repository }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
-          BUILDPACK_ID: ${{ inputs.buildpack_id }}
-          REPO: ${{ github.repository }}
-          RUN_ID: ${{ github.run_id }}
-          SERVER_URL: ${{ github.server_url }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         run: |
-          run_url="${SERVER_URL}/${REPO}/actions/runs/${RUN_ID}"
-          gh pr comment "$PR_NUMBER" --body "📦 Publishing \`${BUILDPACK_ID}\` — workflow started: $run_url"
+          gh pr comment "$PR_NUMBER" --body "Release workflow started: $RUN_URL"
 
   publish:
     name: Publish
@@ -694,10 +706,20 @@ jobs:
           echo "The publish may still complete. Re-run this workflow to check (idempotent recovery)." >> "${GITHUB_STEP_SUMMARY}"
           exit 1
 
+      # Author the Release as the Linguist App. Minted here in the publish job, which already
+      # runs on the IP-allowlisted runner the App token mint requires.
+      - name: Get token for GitHub Release
+        if: steps.preflight.outputs.release_exists != 'true'
+        uses: actions/create-github-app-token@v3
+        id: release-token
+        with:
+          app-id: ${{ inputs.app_id }}
+          private-key: ${{ secrets.app_private_key }}
+
       - name: Create GitHub Release
         if: steps.preflight.outputs.release_exists != 'true'
         env:
-          GH_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ steps.release-token.outputs.token }}
           TAG: ${{ steps.tag.outputs.name }}
           CHANGES: ${{ steps.tag.outputs.changes }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -686,3 +729,56 @@ jobs:
...
+      pull-requests: write     # comment on the triggering prepare-release PR
+    env:
+      MODE: ${{ inputs.qa && 'qa' || 'production' }}
+      SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+    steps:
+      - name: Comment failure on prepare PR
+        if: github.event_name == 'pull_request' && !inputs.qa
+        # Best-effort: a transient PR-comment failure must not abort the job before the more
+        # important production Slack alert (next step) gets a chance to fire.
+        continue-on-error: true
```
edmorley: I think using `continue-on-error` will mean that the overall job will be marked as green even if this step fails, which could obscure any bugs in this step.

It seems an alternative would be to remove `continue-on-error` here, and instead use `if: always()` on the steps after this, per:
https://stackoverflow.com/a/62112985

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -686,3 +729,56 @@ jobs:
...
+    name: Notify Publish Failure
+    needs: [publish]
+    # `failure()` here is scoped to `needs` (the publish job). A skipped publish (its `if:`
+    # evaluated false) yields failure() == false, so no spurious "failed" comment fires.
+    if: failure()
+    runs-on: ubuntu-24.04
+    permissions:
+      pull-requests: write     # comment on the triggering prepare-release PR
+    env:
+      MODE: ${{ inputs.qa && 'qa' || 'production' }}
```
edmorley: Why do we need `MODE`? It's only used once in:
```
if: env.MODE == 'production'
```

...and that could be written using `!inputs.qa` - which would match the usage for the "Comment failure on prepare PR" step (which already uses `!inputs.qa`).

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -32,8 +44,39 @@ defaults:
...
+      github.event_name == 'pull_request'
+      && github.event.pull_request.merged == true
+      && !inputs.qa
+    runs-on: ubuntu-24.04
+    permissions:
+      pull-requests: write     # comment on the triggering prepare-release PR
+    steps:
+      - name: Comment on prepare PR
+        env:
+          GH_TOKEN: ${{ github.token }}
```
edmorley: The prepare release PR will have been created by the Linguist bot's user, rather than the GITHUB_ACTIONS token user (done that way because GitHub Actions skips running workflows for PRs created using GITHUB_ACTIONS to prevent accidental recursive workflows, which would mean our CI wouldn't run otherwise).

However, these comments will then be left by a different user (the GITHUB_ACTIONS user).

Should we use the linguist bot for all actions (such as PR commenting, GitHub Release creation etc) instead?

Switching would also make it consistent with what we do for the CNB automation, eg:
https://github.com/heroku/languages-github-actions/blob/66258117fe7287daee95c0f9f88b0a4a4e2c387b/.github/workflows/_buildpacks-release.yml#L69-L89

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -32,8 +44,39 @@ defaults:
...
+          GH_REPO: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          BUILDPACK_ID: ${{ inputs.buildpack_id }}
+          REPO: ${{ github.repository }}
+          RUN_ID: ${{ github.run_id }}
+          SERVER_URL: ${{ github.server_url }}
+        run: |
+          run_url="${SERVER_URL}/${REPO}/actions/runs/${RUN_ID}"
+          gh pr comment "$PR_NUMBER" --body "📦 Publishing \`${BUILDPACK_ID}\` — workflow started: $run_url"
```
edmorley: I feel like this could be simplified slightly, eg by:
- Skipping BUILDPACK_ID (I don't think we need that in the comment?)
- Concatenating a bunch of the fields in the env vars

eg see how Rune wrote it for the CNB automation's "Notify release started" step here:
https://github.com/heroku/languages-github-actions/blob/66258117fe7287daee95c0f9f88b0a4a4e2c387b/.github/workflows/_buildpacks-release.yml#L69-L89

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -32,8 +44,39 @@ defaults:
     shell: bash
 
 jobs:
+  notify-start:
+    name: Notify Publish Started
```
edmorley: Could you align with the naming used for the CNB automation - so when we're switching between workflows to eg debug things are named similarly between classic and CNBs?
https://github.com/heroku/languages-github-actions/blob/66258117fe7287daee95c0f9f88b0a4a4e2c387b/.github/workflows/_buildpacks-release.yml#L69

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -32,8 +44,39 @@ defaults:
...
+    # Only the merged "Prepare release" PR has a thread to comment on, and that trigger is
+    # always production. The `merged == true` guard mirrors the publish job's universal
+    # precondition so a non-merge pull_request event (opened/synchronize/closed-without-merge)
+    # can't post a false "publishing started" comment while publish itself skips. The explicit
+    # `!inputs.qa` is defensive against any future QA path that might ride a pull_request event.
+    if: |
+      github.event_name == 'pull_request'
+      && github.event.pull_request.merged == true
+      && !inputs.qa
+    runs-on: ubuntu-24.04
```
edmorley: I think this will need to be an IP allowlisted runner (since it's making a write to GitHub, by commenting on the PR), similar to how the CNB release automation does it:
https://github.com/heroku/languages-github-actions/blob/66258117fe7287daee95c0f9f88b0a4a4e2c387b/.github/workflows/_buildpacks-release.yml#L73

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -686,3 +729,56 @@ jobs:
...
             exit 1
           fi
+
+  notify-failure:
+    name: Notify Publish Failure
+    needs: [publish]
+    # `failure()` here is scoped to `needs` (the publish job). A skipped publish (its `if:`
+    # evaluated false) yields failure() == false, so no spurious "failed" comment fires.
+    if: failure()
+    runs-on: ubuntu-24.04
```
edmorley: This will also need to be an IP allowlisted runner

.github/workflows/_classic-buildpack-publish.yml
```diff
@@ -12,10 +12,22 @@ on:
         type: boolean
         required: false
         default: false
+    secrets:
+      SLACK_WEBHOOK_URL:
```
edmorley: I think the convention would be to use lowercase for this secret input, ie: `slack_webhook_url`?

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-of-onworkflow_callsecrets

edmorley (changes requested):
Thank you for adding these!

.github/workflows/_classic-buildpack-publish.yml
colincasey: I considered using Linguist but opted for the GitHub bot since it should be able to perform this job with the least privilege access. Does that change your opinion any @edmorley?

edmorley: @colincasey The bot token is scoped to the single repo, unless you explicitly request more access. The token can also be narrowed further to certain permissions too.

.github/workflows/_classic-buildpack-publish.yml
colincasey: Fixed in 5ee48ac

.github/workflows/_classic-buildpack-publish.yml
colincasey: Fixed in 5ee48ac

.github/workflows/_classic-buildpack-publish.yml
colincasey: Fair enough. Fixed in 5ee48ac

.github/workflows/_classic-buildpack-publish.yml
colincasey: Fixed in 5ee48ac

.github/workflows/_classic-buildpack-publish.yml
colincasey: Fixed in 5ee48ac

.github/workflows/_classic-buildpack-publish.yml
colincasey: Fixed in 5ee48ac

.github/workflows/_classic-buildpack-publish.yml
colincasey: Fixed in 5ee48ac

.github/workflows/_classic-buildpack-publish.yml
colincasey: Fixed in 5ee48ac

.github/workflows/_classic-buildpack-publish.yml
edmorley: @colincasey This doesn't appear to be fixed?

Also, when you update this, could you update the failure step name to match? ("Notify Publish Failure")

edmorley (approved):

.github/workflows/_classic-buildpack-publish.yml
colincasey: 🤦 i confused this change (and the failure step one) with alignment of the comment body.

colincasey: Fixed in [233ffa0](https://github.com/heroku/languages-github-actions/pull/370/commits/233ffa048f4cdf45fb80f74f3795a947eec04e83)

PR #379, opened 2026-07-07
Loosen libcnb version pins so Dependabot can update them

edmorley:
The `libcnb-common`, `libcnb-data` and `libcnb-package` crates pin each other with exact `=` requirements, so when all three are also pinned to an exact version here, Dependabot can't bump any one of them without the others and gives up with "No update possible", opening no PR at all. This is why no `libcnb` 0.31.0 PR was created despite the release.

This only affects this repo and not the CNBs, since this repo depends directly on the lower-level `libcnb-common`/`libcnb-data`/`libcnb-package` crates, whereas the CNBs depend on the higher-level `libcnb`/`libcnb-test`/`libherokubuildpack` crates.

GUS-W-23329202.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,9 +36,12 @@ clap = { version = "4", default-features = false, features = [
 fastrand = "2"
 ignore = "0.4"
 indexmap = "2"
-libcnb-common = "=0.30.4"
-libcnb-data = "=0.30.4"
-libcnb-package = "=0.30.4"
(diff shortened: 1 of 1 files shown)


colincasey (approved):

edmorley:
So it turns out we don't need this after all; merging #380 has fixed Dependabot being able to open PRs for the libcnb group - since #382 has now been opened 🎉

PR #380, opened 2026-07-07
Increase MSRV to Rust 1.96

edmorley:
Bumps the MSRV (`rust-version` in `Cargo.toml`) to the latest stable Rust (1.96), to:

- allow Clippy to enable lints targeting newer Rust
- let the now-MSRV-aware Cargo resolver pick up updates of packages with a recent MSRV
- match the MSRV of our other CNBs/Rust projects

Also refreshes `Cargo.lock`.

GUS-W-23346440.


colincasey (approved):

edmorley:
This also fixed Dependabot being able to update to latest libcnb, as seen by #382 being opened :-)

Issue #383, opened 2026-07-08
Release workflow cache write fails on pull_request trigger after GitHub's read-only cache change


edmorley:
## Summary

The `_buildpacks-release.yml` reusable workflow fails when triggered by its `pull_request: [closed]` event, because the cross-job buildpack cache can no longer be written. The `compile` job logs:

```
Warning: Failed to save: Unable to reserve cache with key 28885892322-compiled-buildpacks. More details: cache write denied: token has no writable scopes
```

The `Cache buildpacks` step (`actions/cache/save`) reports success despite the warning, so no cache is reserved. The `publish-docker` and `publish-github` jobs then fail when restoring that cache with `fail-on-cache-miss: true`.

Example failure (buildpacks-procfile v4.2.3):
https://github.com/heroku/buildpacks-procfile/actions/runs/28885892322/job/85737625538

## Cause

On 2026-06-26 GitHub shipped [Read-only Actions cache for untrusted triggers](https://github.blog/changelog/2026-06-26-read-only-actions-cache-for-untrusted-triggers/), which issues read-only cache tokens for the default-branch scope on `pull_request`-triggered runs. Our release workflow is triggered by `pull_request` (activity type `closed`), so its cache token no longer has writable scopes.

Confirmed by comparing two runs of the identical workflow (same `permissions: {}`, both printing `GITHUB_TOKEN Permissions: Metadata: read`):

| | Event | Ref | `Cache buildpacks` |
|---|---|---|---|
| Last success (2026-06-25) | `pull_request` | `prepare-release` | ✅ saved |
| Failure (2026-07-07) | `pull_request` | PR merge ref | ❌ `token has no writable scopes` |
| Manual recovery (2026-07-07) | `workflow_dispatch` | `main` | ✅ saved |

`push`, `workflow_dispatch`, `schedule`, etc. remain trusted (read-write cache), which is why a manual `workflow_dispatch` on `main` succeeds. Re-running the failed `pull_request` run does not help — it replays the same untrusted event context.

This affects every repo consuming this reusable workflow (buildpacks-procfile, -go, -php, -deb-packages, -dotnet, -ruby, -nodejs, …), so all releases via the auto-merge path are currently broken.

## Why the failure surfaced late (and was hard to debug)

[...]

PR #386, opened 2026-07-14
Grant actions: write scope to compile job for cache/save@v6

Malax:
Every release fails at cache handoff. The caller workflows set `permissions: {}`, and `actions/cache/save@v6` needs the `actions: write` scope to write to the cache. The compile job's save fails with `cache write denied: token has no writable scopes`, and every downstream publish job fails restore with `fail-on-cache-miss`.

This came in with #374 (v5 → v6). v5 relied on the legacy cache backend which did not require the scope; v6 fully depends on the v2 backend which does.

Scope is granted only to the `compile` job. Restore does not need write, so every publish job continues to inherit the top-level `permissions: {}`.

Longer term we should probably migrate the intra-run buildpack handoff from `actions/cache` to `actions/upload-artifact` / `download-artifact`, which is the intended primitive for passing files between jobs in the same run and needs no elevated scope. That's a bigger change though — this PR just aims to restore our ability to release.

Example failure: https://github.com/heroku/buildpacks-jvm/actions/runs/29342731996

### DIFF
--- a/.github/workflows/_buildpacks-release.yml
+++ b/.github/workflows/_buildpacks-release.yml
@@ -91,6 +91,11 @@ jobs:
   compile:
     name: Compile Buildpacks
     runs-on: ubuntu-24.04
+    # Required by actions/cache/save@v6: writing to the GitHub Actions cache
+    # requires the `actions: write` scope on the workflow token. Reading from
+    # the cache does not, so publish jobs keep the top-level `permissions: {}`.
+    permissions:
+      actions: write
     outputs:
       buildpacks: ${{ steps.generate-buildpack-matrix.outputs.buildpacks }}
       version: ${{ steps.generate-buildpack-matrix.outputs.version }}


.github/workflows/_buildpacks-release.yml
```diff
@@ -91,6 +91,11 @@ jobs:
   compile:
     name: Compile Buildpacks
     runs-on: ubuntu-24.04
+    # Required by actions/cache/save@v6: writing to the GitHub Actions cache
+    # requires the `actions: write` scope on the workflow token. Reading from
+    # the cache does not, so publish jobs keep the top-level `permissions: {}`.
+    permissions:
+      actions: write
```
edmorley, line 98: Are you sure this is the real cause? The PR description mentions the cache action v6 upgrade, but that only contained an ESM upgrade. Did you see https://github.com/heroku/languages-github-actions/issues/383 ?

Malax: Nope, not sure. I didn't see #383 yet - I'll check.

Malax:
Closing, ref: https://github.com/heroku/languages-github-actions/issues/383

